PR #3, opened 2015-06-12
Collect static assets for production

SteveMarshall:
Jenkins needs to be able to extract assets from the container when deploying. To allow it to do that, compile them then collect them all into one place (defined by `mtp_cashbook.settings.*.STATIC_ROOT`).

Because of the way we're deploying them to S3, `STATIC_ROOT` needs to be the same as `STATIC_URL`, or we'll be referring to invalid URLs.

Also use `npm install` instead of `bower install` so that it installs gulp et al. in more accessible paths.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,10 @@ ADD ./requirements/ /app/requirements/
 RUN pip3 install -r requirements/prod.txt
 
 ADD . /app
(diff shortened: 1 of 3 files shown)


marcofucci:
:+1:

r4vi:
:+1:

PR #4, opened 2015-06-12
Don't set default timezone to be Europe/London

r4vi:
After discussing with platofrms team, this isn't a good idea. Their own
docker base images don't do this and we want to be consistent.


marcofucci:
:+1:

PR #5, opened 2015-06-12
Make running a local dev environment quicker

r4vi:
After setting up a new dev environment for the API server project
I have found some better ways of doing things in a local dev
environment using docker-compose.

 - The docker-compose up command now installed python dev
   dependencies before running the server
 - The npm install command in the gulp container now shares
   a .npm cache with the host
 - PIP now shares a PIP_DOWNLOAD_CACHE with the host so it
   doesn't need to download the same dependencies every
   time

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,16 +1,23 @@
 django:
   build: .
-  command: python manage.py runserver 0.0.0.0:8001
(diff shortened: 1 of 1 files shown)


marcofucci:
:+1:

PR #11, opened 2015-06-23
Allow users to log into cashbook

marcofucci:
This allows a valid user to log into the cashbook app using oauth2 against the API app.

Custom login, logout and authentication logic is required as the standard django behaviour
relies on database tables.
Unfortunately, I had to borrow some django code and change it accordingly in order to support our authentication method.

Use the mock python library for mocking the API connection so that unit tests can run
without dependencies.

### DIFF
--- a/mtp_cashbook/apps/mtp_auth/exceptions.py
+++ b/mtp_cashbook/apps/mtp_auth/exceptions.py
@@ -1,0 +1,6 @@
+class ConnectionError(Exception):
+    """
+    Exception that should be raised in case of connection
+    problems with the api server.
+    """
+    pass
--- a/mtp_cashbook/apps/mtp_auth/forms.py
+++ b/mtp_cashbook/apps/mtp_auth/forms.py
@@ -1,0 +1,53 @@
+from django import forms
+from django.contrib.auth import authenticate
+
+from django.utils.translation import ugettext_lazy as _
+
+from .exceptions import ConnectionError
+
+
+class AuthenticationForm(forms.Form):
+    """
+    Authentication form used for authenticating users during the login process.
+    """
+    username = forms.CharField(label=_("Username"), max_length=254)
+    password = forms.CharField(label=_("Password"), widget=forms.PasswordInput)
+
+    error_messages = {
+        'invalid_login': _("Please enter a correct username and password. "
+                           "Note that both fields may be case-sensitive."),
+        'connection_error': _("The API Server seems down, please try again later."),
+    }
+
+    def __init__(self, request=None, *args, **kwargs):
+        self.request = request
+        self.user_cache = None
+        super(AuthenticationForm, self).__init__(*args, **kwargs)
+
+    def clean(self):
+        username = self.cleaned_data.get('username')
+        password = self.cleaned_data.get('password')
+
+        if username and password:
+            try:
+                self.user_cache = authenticate(
+                    username=username, password=password
+                )
+                # if authenticate returns None it means that the
+                # credentials were wront
+                if self.user_cache is None:
+                    raise forms.ValidationError(
+                        self.error_messages['invalid_login'],
+                        code='invalid_login',
+                    )
+            except ConnectionError:
+                # in case of problems connecting to the api server
+                raise forms.ValidationError(
+                        self.error_messages['connection_error'],
+                        code='connection_error',
+                    )
+
+        return self.cleaned_data
+
+    def get_user(self):
+        return self.user_cache
--- a/mtp_cashbook/apps/mtp_auth/middleware.py
+++ b/mtp_cashbook/apps/mtp_auth/middleware.py
@@ -1,0 +1,23 @@
+from django.utils.functional import SimpleLazyObject
+
+from . import get_user as auth_get_user
+
+
+def get_user(request):
+    """
+    Returns a cached copy of the user if it exists or calls `auth_get_user`
+    otherwise.
+    """
+    if not hasattr(request, '_cached_user'):
+        request._cached_user = auth_get_user(request)
+    return request._cached_user
+
+
+class AuthenticationMiddleware(object):
+    """
+    It simply sets `request.user` so that it can be used in our views.
+
+    The build-in Django one sadly tries to get the user from the database.
+    """
+    def process_request(self, request):
+        request.user = SimpleLazyObject(lambda: get_user(request))
--- a/mtp_cashbook/apps/mtp_auth/models.py
+++ b/mtp_cashbook/apps/mtp_auth/models.py
@@ -1,0 +1,29 @@
+class MtpUser(object):
+    """
+    Authenticated user, similar to the Django one.
+
+    The built-in Django `AbstractBaseUser` sadly depends on a few tables and
+    cannot be used without a datbase so we had to create a custom one.
+    """
+    def __init__(self, token):
+        self.pk = token
+        self.is_active = True
+
+    def save(self, *args, **kwargs):
+        pass
+
+    def is_authenticated(self, *args, **kwargs):
+        return True
+
+
+class MtpAnonymousUser(object):
+    """
+    Anonymous non-authenticated user, similar to the Django one.
+
+    The built-in Django `AnonymousUser` sadly depends on a few tables and
+    gives several warnings when used without a database so we had to create a
+    custom one.
+    """
+
+    def is_authenticated(self, *args, **kwargs):
+        return False
--- a/mtp_cashbook/apps/mtp_auth/views.py
+++ b/mtp_cashbook/apps/mtp_auth/views.py
@@ -1,0 +1,101 @@
+from django.conf import settings
+from django.contrib.auth import REDIRECT_FIELD_NAME
+from django.contrib.sites.shortcuts import get_current_site
+from django.http import HttpResponseRedirect
+from django.shortcuts import resolve_url
+from django.template.response import TemplateResponse
+from django.utils.http import is_safe_url
+from django.utils.translation import ugettext as _
+from django.views.decorators.cache import never_cache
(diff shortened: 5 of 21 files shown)


marcofucci:
@r4vi @SteveMarshall do you want to have a look?

mtp_cashbook/apps/mtp_auth/exceptions.py
```diff
@@ -0,0 +1,6 @@
+class ConnectionError(Exception):
+    """
+    Exception that should be raised in case of connection
+    problems with the api server.
+    """
+    pass
```
SteveMarshall: Do we need our own `ConnectionError` class? Isn't this a little confusing, given that [Python 3 provides a `ConnectionError`](https://docs.python.org/3/library/exceptions.html#ConnectionError)? Should we just use that (or make this a subclass… or just pass out whatever the underlying error is)?

mtp_cashbook/apps/mtp_auth/forms.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Authentication form used for authenticating users during the login process.
+    """
+    username = forms.CharField(label=_("Username"), max_length=254)
+    password = forms.CharField(label=_("Password"), widget=forms.PasswordInput)
+
+    error_messages = {
+        'invalid_login': _("Please enter a correct username and password. "
+                           "Note that both fields may be case-sensitive."),
+        'connection_error': _("The API Server seems down, please try again later."),
```
SteveMarshall: I'd be tempted to explicitly make these translation keywords, rather than the literal english text we expect them to be. That way, when @leighmoney (or whoever) wants to improve the copy, we don't have to dig around in the code to make that change.

SteveMarshall: That would also mean that we can get rid of the `error_messages` hash here and just use `_()` directly through the code. I'm not sure if I prefer that or not, mind you.

mtp_cashbook/apps/mtp_auth/forms.py
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    def __init__(self, request=None, *args, **kwargs):
+        self.request = request
+        self.user_cache = None
```
SteveMarshall: Why `user_cache`? Unless I'm missing something, this feels more like `current_user`?

mtp_cashbook/apps/mtp_auth/forms.py
```diff
@@ -0,0 +1,53 @@
...
+        super(AuthenticationForm, self).__init__(*args, **kwargs)
+
+    def clean(self):
```
SteveMarshall: This doesn't feel like `clean` to me: it seems like this it's getting the cleaned inputs and then doing the actual login… Oh, wait, is this a holdover of this being a subclass of `django.forms.Form`? If so, it might be nice to break this out more clearly into a `login` method that the `clean` method calls, perhaps?

mtp_cashbook/apps/mtp_auth/forms.py
SteveMarshall: This might also reduce the need for the comment about authenticate returning `None` below.

mtp_cashbook/apps/mtp_auth/exceptions.py
marcofucci: Yes I knew that you might have not liked this.
It's not confusing for me as I wanted to make explicit that you need to catch **this** exception and understand what it means for you as dev when you do it.
If it's confusing for you, I might just use the requests one or the python3 one, doesn't really change anything for me really.

mtp_cashbook/apps/mtp_auth/middleware.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def get_user(request):
+    """
+    Returns a cached copy of the user if it exists or calls `auth_get_user`
+    otherwise.
+    """
+    if not hasattr(request, '_cached_user'):
+        request._cached_user = auth_get_user(request)
+    return request._cached_user
```
SteveMarshall: Again: `_cached_user` might be clearer as `_current_user` unless there's some nuance I'm missing?

mtp_cashbook/apps/mtp_auth/models.py
```diff
@@ -0,0 +1,29 @@
+class MtpUser(object):
+    """
+    Authenticated user, similar to the Django one.
+
+    The built-in Django `AbstractBaseUser` sadly depends on a few tables and
+    cannot be used without a datbase so we had to create a custom one.
```
SteveMarshall: This stuff would make perfect material for a commit message rather than a comment, if this were done in multiple discreet commits :wink:

mtp_cashbook/apps/mtp_auth/forms.py
marcofucci: Those are just keys as any other keys, you will be able to change them in the translation files if you want although I doubt we need translation files.

This is standard django convention, explicit english translation keys and error_messages hash.
I suggest we shouldn't change it as it would be too confusing for devs.

mtp_cashbook/apps/mtp_auth/forms.py
marcofucci: Wanted to keep it the same as django which uses this name, I don't see any problems really

mtp_cashbook/templates/base.html
```diff
@@ -21,7 +21,7 @@
   {% endblock content %}
 
   {% if DEBUG %}
-  
+
   <script src="{% static "javascripts/app.js" %}"></script>
   <script src="//localhost:35729/livereload.js"></script>
```
r4vi: not part of this PR but this needs to probably be deleted

r4vi:
Looks good to me :+1:

mtp_cashbook/apps/mtp_auth/views.py
```diff
@@ -0,0 +1,101 @@
...
+        request.GET.get(redirect_field_name, '')
+    )
+
+    if request.method == "POST":
+        form = authentication_form(request, data=request.POST)
+        if form.is_valid():
+
+            # Ensure the user-originating redirection url is safe.
+            if not is_safe_url(url=redirect_to, host=request.get_host()):
+                redirect_to = resolve_url(settings.LOGIN_REDIRECT_URL)
```
SteveMarshall: :+1:

mtp_cashbook/apps/mtp_auth/forms.py
marcofucci: yes, it's the django `clean` method where this logic should be.

I think the code is clear enough for django devs. If you have a different `login` method, it wouldn't be clear that you can't call it before calling clean so you would expect to call `login` and just work.

mtp_cashbook/templates/mtp_auth/login.html
```diff
@@ -0,0 +1,39 @@
+{% extends "base.html" %}
+
+{% block content %}
+{% if form.errors and not form.non_field_errors %}
+<p class="errornote">
+{% if form.errors.items|length == 1 %}Please correct the error below.{% else %}Please correct the errors below{% endif %}
```
SteveMarshall: Can we wrap this for translation?

mtp_cashbook/apps/mtp_auth/forms.py
SteveMarshall: This is the standard in Django? Huh. Well, let's stick with this then, though the English-as-default thing always makes me uncomfortable.

mtp_cashbook/apps/mtp_auth/forms.py
SteveMarshall: If this is following Django convention, then OK!

mtp_cashbook/apps/mtp_auth/forms.py
marcofucci: well, it's not english as default but more "A string like this" is better than A_STRING_LIKE_THIS.
Does it make more sense?

SteveMarshall: I'm not sure I buy that argument, but that's one for the pub :beers: Leave it as it is, for the reasons you say.

mtp_cashbook/apps/mtp_auth/middleware.py
marcofucci: Again, that's how django called that variable so I don't want to change it unless it's really necessary.

mtp_cashbook/templates/mtp_auth/login.html
marcofucci: Do we need translation?

mtp_cashbook/apps/mtp_auth/forms.py
SteveMarshall: Having had a poke around Django I see that, yeah, you're right. I'll shut up :stuck_out_tongue:

mtp_cashbook/templates/mtp_auth/login.html
SteveMarshall: Well, we're doing it elsewhere, and we'll need to support Welsh at some point so if it's easy-ish, we should do it.

mtp_cashbook/apps/mtp_auth/middleware.py
SteveMarshall: Yup, fine. To make a point: we wouldn't be having this discussion if this were in a discrete commit explaining why we're rolling our own and that we're basically mirroring the Django core auth with small changes to support oauth.

mtp_cashbook/apps/mtp_auth/exceptions.py
SteveMarshall: No, that's fine by me.

PR #12, opened 2015-06-25
Apply GOV.UK styling using Mojular framework

matharden:
This is pretty much "[Style home page summary boxes](https://github.com/ministryofjustice/money-to-prisoners/pull/9)" and "[HTML template and internal styling](https://github.com/ministryofjustice/money-to-prisoners/pull/7)" for Django.

### DIFF
--- a/mtp_cashbook/templates/core/index.html
+++ b/mtp_cashbook/templates/core/index.html
@@ -1,14 +1,22 @@
 {% extends "base.html" %}
 
-{% block title %}Yo{% endblock title %}
-
 {% block content %}
-<div class="row">
-  <div class="small-12 columns moj-wrapper">
-    <h1>
-      You are logged in!
-    </h1>
-    <p><a href="{% url 'auth:logout' %}">Log out</a></p>
-  </div>
-</div>
+  <h1>Money sent to prisoners</h1>
+
+  <ul class="SummaryBoxes">
+    <li class="SummaryBoxes--unprocessed">
+      <a href="#TODO">
+        <h2>New</h2>
+        <p><strong>54</strong> Payments to credit</p>
+      </a>
+    </li>
+
+    <li>
+      <a href="#TODO">
+        <h2>History</h2>
+        <p><span>Payments</span> credited by you</p>
+      </a>
+    </li>
+  </ul>
+
 {% endblock content %}
--- a/mtp_cashbook/templates/core/index.html
+++ b/mtp_cashbook/templates/core/index.html
@@ -1,20 +1,29 @@
 {% extends "base.html" %}
+{% load i18n %}
 
 {% block content %}
-  <h1>Money sent to prisoners</h1>
+  <h1>{% trans "Money sent to prisoners" %}</h1>
 
   <ul class="SummaryBoxes">
     <li class="SummaryBoxes--unprocessed">
       <a href="#TODO">
-        <h2>New</h2>
-        <p><strong>54</strong> Payments to credit</p>
+        <h2>{% trans "New" %}</h2>
+        <p>
+          {% blocktrans with payments_count="<strong>54</strong>" %}
+          {{ payments_count }} Payments to credit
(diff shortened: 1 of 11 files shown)


mtp_cashbook/templates/core/index.html
```diff
@@ -1,15 +1,22 @@
...
-      Yo MoJ222!&nbsp;
-      <a href="//github.com/mixxorz/generator-moj-django/"><i class="fa fa-github"></i></a>
-    </h1>
-    <p>Welcome to your MoJ Django app.</p>
-  </div>
-</div>
+  <h1>Money sent to prisoners</h1>
+
+  <ul class="SummaryBoxes">
+    <li class="SummaryBoxes--fresh">
```
SteveMarshall: What does `--fresh` indicate?

Also, the duplication of `SummaryBoxes` here seems odd to me; why not use the cascade (e.g. `.SummaryBoxes .fresh`)?

Also also: why CamelCase? It's harder to read and, if memory serves, behaves inconsistently in DOM implementations (though that may only be the case for super-old IE versions…)?

SteveMarshall:
I wonder if it's worth wrapping the strings in [translation markers](https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#trans-template-tag) just to start as we mean to go on… Or at least the ones that aren't clearly placeholders that're going to be changed imminently? What about wrapping those, and marking the ones that need changing in a `needs translation` comment till they're ready?

SteveMarshall:
Might also be nice to have slightly more descriptive commit messages. For example, the first might mention *why* you're using Mojular, and why your making the changes to the Django/Gulp files, perhaps?

mtp_cashbook/templates/core/index.html
matharden: `.SummaryBoxes--fresh` denotes a variation (hence `--`) of a `li` in the `.SummaryBoxes` module. It's a naming convention base on [SUIT CSS](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md), which is similar but not as dogmatic as "BEM". I'm using this as a prefix to any sub elements which require a class.

The CamelCase is a convention I am using which provides an easy way to identify the root element of HTML modules. This will become clearer as I create more modules. I will check this with the browsers we need to support.

matharden: And the `--fresh` bit is my stupid name for the outstanding payments variation of the summary boxes. Maybe `--unprocessed` would be a better variation classname?

SteveMarshall: Yeah, `--unprocessed` would be clearer, I think. As discussed elsewhere, I'm happy to roll with the rest and trust your judgement.

marcofucci:
Retest this please

marcofucci:
:+1:

mtp_cashbook/templates/core/index.html
```diff
@@ -1,14 +1,22 @@
...
+        <h2>New</h2>
+        <p><strong>54</strong> Payments to credit</p>
+      </a>
+    </li>
+
+    <li>
+      <a href="#TODO">
+        <h2>History</h2>
+        <p><span>Payments</span> credited by you</p>
+      </a>
```
SteveMarshall: Be nice to translate these, even though we don't think we need other languages yet, because it'll make @leighmoney's life easier later.

Rough examples of how I'd do these blocks, as discussed (though no guarantees that this exact code will work):

```
{% blocktrans with payments_to_credit="<strong>{{ payments }}</strong>"  %}
{{ payments_to_credit }} payments
{% endblocktrans %}

{% blocktrans with highlight="<span>" endhighlight="</span>" %}
{{ highlight }}Payments{{ endhighlight }} credited by you
{% endblocktrans %}
```

matharden: You say "rough", I say "copy/paste".

PR #13, opened 2015-06-25
Pass the host's API_URL to the cashbook container

SteveMarshall:
When running in docker-compose with boot2docker, the cashbook attempts to connect to the default API URL (`http://localhost:8000`) but fails because `localhost` refers to the current container.

To resolve that, we need to pass the `API_URL` environment variable with the boot2docker VM's IP into the container. By specifying `API_URL` in `docker-compose.yml` with no value, docker-compose will read the value from the host, if it exists, and pass that to the container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,13 @@ In a terminal `cd` into the directory you checked this project out into, then
 $ docker-compose build && docker-compose up
 ```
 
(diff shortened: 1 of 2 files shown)


marcofucci:
:+1:

marcofucci:
@r4vi you can merge it if you're happy with it

PR #14, opened 2015-06-26
Refactor api_client to support refresh tokens automatically.

marcofucci:
This changes the way the code calls the api simplifying its logic.
Only 2 functions are exposed:
- `api_client.authenticate` which authenticates against the api
- `api_client.get_connection` which returns a slumber connection able
to deal with refreshing the access token automatically if needed.

In this way, the views can just use the connection without worrying about
tokens at all.


marcofucci:
@r4vi @SteveMarshall OK?

r4vi:
:+1: looks good

SteveMarshall:
Retest this please

PR #18, opened 2015-07-06
Make TransactionView use real data instead of hardcoded values 

r4vi:
@matharden I've moved a few things around but this is how you get real data


mtp_cashbook/templates/cashbook/transaction_list.html
```diff
@@ -27,14 +28,17 @@
...
+      <tr {% if forloop.counter|divisibleby:2 %}{% else %}class="odd"{% endif %}>
+        <td>{{ transaction.prisoner_number }}</td>
+        <td>??? ???</td>
+        <td>&pound;{{ transaction.amount|currency }}</td>
+        <td>{{ transaction.sender }}</td>
+        <td class="check"><input type="checkbox" {% if transaction.credited %}checked{% endif %}></td>
       </tr>
-
+    {% empty %}
+      {% trans "No transactions found" %}
```
mattemp: I'll change this later, as this will break the markup.

mtp_cashbook/apps/cashbook/views.py
```diff
@@ -0,0 +1,23 @@
+from django.contrib.auth.decorators import login_required
+from django.utils.decorators import method_decorator
+from django.views.generic import ListView
+from mtp_cashbook.apps.mtp_auth.api_client import get_connection
+
+
+class TransactionListView(ListView):
```
mattemp: Could we rename this to something which indicates it's a batch of payments rather than all please?

matharden:
:+1:

PR #19, opened 2015-07-06
Add user data to User Model

marcofucci:
This makes an extra query to `/users/<username>/` during the authentication process so that request.user and its user data can be cached and accessed by views/templates.

### DIFF
--- a/mtp_cashbook/apps/mtp_auth/models.py
+++ b/mtp_cashbook/apps/mtp_auth/models.py
@@ -6,18 +6,40 @@ class MtpUser(object):
     cannot be used without a datbase so we had to create a custom one.
     """
 
-    def __init__(self, username, token={}):
-        self.pk = username
-        self.username = username
+    def __init__(self, pk, token, user_data):
+        self.pk = pk
         self.is_active = True
 
         self.token = token
+        self.user_data = user_data
 
     def save(self, *args, **kwargs):
         pass
 
     def is_authenticated(self, *args, **kwargs):
         return True
+
+    @property
+    def username(self):
+        return self.user_data.get('username')
+
+    @property
+    def get_full_name(self):
+        if not hasattr(self, '_full_name'):
+            name_parts = [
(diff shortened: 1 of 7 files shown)


marcofucci:
@r4vi @SteveMarshall review/merge ?

mtp_cashbook/templates/base.html
```diff
@@ -21,10 +21,12 @@
 {% endblock %}
 
 {% block global_subheader %}
+  {% if request.user.is_authenticated %}
   <div class="user-details">
-    {% trans "Logged in as Firstname Lastname" %}
+    {% trans "Logged in as " %}{{ request.user.get_full_name|default:request.user.username }}
```
SteveMarshall: This might be better as:

```
{% blocktrans with username=request.user.get_full_name|default:request.user.username %}
Logged in as {{ username }}.
{% endblocktrans %}
```
… or similar. That way the translation files will have better context for editing.

mtp_cashbook/apps/mtp_auth/models.py
```diff
@@ -19,6 +19,28 @@ def save(self, *args, **kwargs):
...
+                self.user_data.get('first_name'),
+                self.user_data.get('last_name')
+            ]
+            self._full_name = ' '.join(filter(None, name_parts))
+        return self._full_name
+
+    @property
+    def prison(self):
+        try:
+            return self.user_data.get('prisons', [])[0]
```
SteveMarshall: Would it be better to return the entire array of prisons the user has access to here, enabling something like `if prison in user.prisons`? That'd also remove the need for the `[0]`, I think?

marcofucci: The cashbook doesn't know which prisons the user has access to, so you access this prop to get that data. We support only one prison at the moment so I was thinking to add a warning when authenticating. Something like: `if prisons.length > 1 => log warning`. Not sure if it helps though.

PR #22, opened 2015-07-08
Implement credit/discard view

marcofucci:
This changes the transaction batch view to being a FormView so that we can
validate and manage the it easily.
Saving the form will credit the selected transactions and discard the rest.
Saving the form passing discard='1' will discard all the transactions.


marcofucci:
@SteveMarshall @matharden a review/merge pls?

mtp_cashbook/apps/cashbook/forms.py
```diff
@@ -0,0 +1,71 @@
+from django import forms
+from django.utils.functional import cached_property
+
+from mtp_cashbook.apps.mtp_auth.api_client import get_connection
+
+
+class ProcessTransactionBatchForm(forms.Form):
+    transactions = forms.MultipleChoiceField(choices=(), required=False)
```
SteveMarshall: This is a really neat way of representing the list of transactions :+1:

mtp_cashbook/apps/cashbook/forms.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def save(self):
+        all_ids = {str(t[0]) for t in self.transaction_choices}
+        to_credit = set(self.cleaned_data['transactions'])
+        to_discard = all_ids - to_credit
+
+        # The two calls to the api might return a status code != 2xx
+        # but we are allowing them to raise an exception here.
+        # This shouldn't really happen but if it does often enough,
+        # we might want to add some exception handling.
```
SteveMarshall: As discussed, I've [added a task next sprint for this](https://www.pivotaltracker.com/story/show/98643080).

mtp_cashbook/apps/cashbook/tests/test_forms.py
```diff
@@ -0,0 +1,122 @@
...
+        ] + [self.pending_transactions_response]
+
+        form = ProcessTransactionBatchForm(self.request)
+        expected_choices = [
+            (t['id'], t) for t in self.pending_transactions_response['results']
+        ]
+        self.assertEqual(form.transaction_choices, expected_choices)
+        self.assertEqual(form.fields['transactions'].choices, expected_choices)
+
+    def test_credit_some(self):
```
SteveMarshall: We test crediting some, and discarding all; is it worth testing crediting all, too?

mtp_cashbook/apps/cashbook/tests/test_forms.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+
+        # checking call to release
+        mocked_release = form.client.transactions()().release.post
+        self.assertEqual(mocked_release.call_count, 1)
+        discarded_transactions = mocked_release.call_args[0][0]
+        self.assertListEqual(
+            sorted(discarded_transactions['transaction_ids']),
+            sorted(['1', '4'])
+        )
```
SteveMarshall: Could we break this into two tests? That way, it's clearer when one or other action breaks… For example, something like `test_can_credit_some_transactions` and then `test_release_after_partial_credit_only_releases_uncredited_transactions` (which is a really awful name, but you get the point).

mtp_cashbook/apps/cashbook/views.py
```diff
@@ -1,27 +1,47 @@
...
         return context
+
+    def form_valid(self, form):
+        credited, discarded = form.save()
+
+        messages.success(
+            self.request,
+            _(
+                'Batch processed successfully. \
+                %(credited)s credited, %(discarded)s discarded.'
```
SteveMarshall: Hm. We use `%()` here and `{{}}` in templates. I wonder if we can make those consistent to make copy-editing easier?

mtp_cashbook/apps/cashbook/tests/test_forms.py
marcofucci: It would be really confusing for me to have different tests as we are testing one global thing: given a set of ids, some should get credited and the rest should get discarded.

Also, `test_release_after_partial_credit_only_releases_uncredited_transactions` would have to check that the credit bit worked anyway so it wouldn't be better than this.
Also if the credit bit broke `test_release_after_partial_credit_only_releases_uncredited_transactions` would break and it would be even harder to understand what happened.

mtp_cashbook/apps/cashbook/views.py
marcofucci: you use `%()` in python code and `blocktrans` +  `{{}}` in templates and django does everything for you. You'll end up with `%()` in your translation files in both cases.

SteveMarshall: Oh, cool: fine by me, then! :+1:

mtp_cashbook/apps/cashbook/tests/test_forms.py
marcofucci: `test_credit_all` added

mtp_cashbook/apps/cashbook/tests/test_forms.py
SteveMarshall: It's just dawned on me that I'm reading the test wrong: I was reading the `checking call to patch` and `checking call to release` bits as if they were testing distinct bits of behaviour when, in fact, they're validating that the `form.save()` above behaved as expected. Ignore me! :weary:

SteveMarshall:
Looks good to me; @matharden, this is going into your branch, so can you review/merge?

matharden:
This branch needs to be rebased again `:batch`, which is now in a PR (#23). With that in mind, if might be easier to wait until that is merged and PR this against master.

matharden:
@marcofucci I got a `Forbidden (403) CSRF verification failed. Request aborted.` error when trying to submit or discard.

marcofucci:
@matharden the problem somehow happens when using gulp serve and runserver at the same time, that is, when you log into `localhost:8001` and `localhost:3000`. The cookies are shared and django or python doesn't like that.

To fix that, just remove all the cookies for the domain localhost and it should work fine.

marcofucci:
Will rebase and PR from master

PR #24, opened 2015-07-09
Stop docker-compose building our image twice

SteveMarshall:
The djangogulpserve container uses the same image as the django container, so we should explicitly indicate that's what we want to avoid building it twice. If djangogulpserve needed to be run on its own
this would break, but we never actually need that because it only exists for serving assets during development.

Also make the the environment variable passed to `--port` consistent with the one used for `--host`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,8 +14,13 @@ django:
     API_URL:
 
 djangogulpserve:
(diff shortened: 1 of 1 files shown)


marcofucci:
:+1:

PR #25, opened 2015-07-09
Automate the getting started process

SteveMarshall:
There are a handful of complex, conditional steps people have to jump through before being able to work with our codebase. This attempts to automate them all using `make`, including starting `boot2docker` (if it's installed; otherwise we assume it's unnecessary), and means we can simplify the process of working with our code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,41 @@
-ifneq ($(shell command -v boot2docker),)
-API_ENDPOINT=API_URL=http://`boot2docker ip`:8000
-endif
+ERROR=\033[0;31m   # Red
+WARNING=\033[0;33m # Yellow
+RESET=\033[0m
+err=echo "$(ERROR)$(1)$(RESET)"
+warn=echo "$(WARNING)$(1)$(RESET)"
 
 run: build
 	$(API_ENDPOINT) docker-compose up
 
 build: .dev_django_container
 
-.dev_django_container: .
+ifneq ($(shell command -v boot2docker),)
+API_ENDPOINT=API_URL=http://`boot2docker ip`:8000
+
+b2d_status=$(shell boot2docker status)
+
+ifneq ($(b2d_status),running)
+boot2docker_up=boot2docker_up
+boot2docker_up:
+	boot2docker init
+	boot2docker up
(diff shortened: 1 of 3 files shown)


SteveMarshall:
@matharden @marcofucci What d'you guys think of this? There are a few other things I'd like to add to it (notably, running tests for both Python and Javascript), so don't merge just yet…

marcofucci:
:+1: if that helps devs it's fine, I don't use docker-compose that so can't comment on it.
I guess we add it and see if people like it?

SteveMarshall:
I think this adds enough value as-is to be merged. @matharden waddya think? Anything need changing/not clear?

README.md
```diff
@@ -3,41 +3,38 @@ The Cashbook UI for the Money to Prisoners Project
 
 ## Dependencies
 ### Docker
-To run this project locally you need to have 
-[Docker](http://docs.docker.com/installation/mac/) and 
+To run this project locally you need to have
+[Docker](http://docs.docker.com/installation/mac/)and
```
matharden: Needs space before "and"

SteveMarshall: Well spotted!

matharden:
I think that's much easier to run. Nice work!

Makefile
```diff
@@ -0,0 +1,53 @@
...
+	    "pip install --quiet -r requirements/dev.txt && \
+	     /app/manage.py test $(TEST)"
+
+run: build
+	$(API_ENDPOINT) docker-compose up
+
+build: .dev_django_container
+
+ifneq ($(shell command -v boot2docker),)
+API_ENDPOINT=API_URL=http://`boot2docker ip`:8000
```
matharden: Could[/should] this be indented from here to the `endif`?

SteveMarshall:
@matharden @marcofucci I'm pretty sure we shouldn't do 03be1e9, but have a look and let me know if you think it's as crazy as I do…

SteveMarshall:
After discussion with @marcofucci, I've rolled back so we're not including 03be1e9 now. It's an interesting diversion, but has too many unexpected side effects.

PR #26, opened 2015-07-09
Implement credit/discard view

marcofucci:
This changes the transaction batch view to being a FormView so that we can
validate and manage the it easily.
Saving the form will credit the selected transactions and discard the rest.
Saving the form passing discard='1' will discard all the transactions.


marcofucci:
@SteveMarshall same as before but with an added commit: 26f2404d3c380c922a4fcf55e6e852561a3454a0

PR #29, opened 2015-07-10
Add linting helper and fix existing linting issues

SteveMarshall:
We need an easy way of checking that our code matches [pep8](https://www.python.org/dev/peps/pep-0008/) (Python's style guide). This adds `make lint` for that purpose.

Also fixes linting errors in the current codebase.

It's possibly worth reviewing this one commit at a time. Once it's merged to master, I'll look at linting in Jenkins.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ node_modules
 
 mtp_cashbook/assets-src/bower_components
 mtp_cashbook/assets/
+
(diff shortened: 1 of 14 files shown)


SteveMarshall:
Argh, just realised I need to rebase this once #25 is merged to master. Ignore the first few commits!

PR #30, opened 2015-07-14
Add linting helper and fix existing linting issues

SteveMarshall:
We need an easy way of checking that our code matches [pep8](https://www.python.org/dev/peps/pep-0008/) (Python's style guide). This adds `make lint` for that purpose, and fixes existing linting errors.

It's probably worth reviewing this one commit at a time. Once it's merged to master, I'll look at linting in Jenkins.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ all: lint test run
 lint: build
 	docker-compose run django bash -c \
 	    "pip install --quiet -r requirements/dev.txt && \
-	     flake8 ."
+	     flake8 $(LINT_OPTS) ."
 
 test: build
 	$(API_ENDPOINT) docker-compose run django bash -c \
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,16 @@ all: lint test run
 lint: build
 	docker-compose run django bash -c \
 	    "pip install --quiet -r requirements/dev.txt && \
-	     flake8 $(LINT_OPTS) ."
+	     flake8 $(LINT_OPTS) . && \
+	     find . \( -path ./node_modules -o \
+	               -path ./mtp_cashbook/assets-src/bower_components -o \
+	               -path ./requirements \
+	            \) \
+	            -prune -o \( \
+	               -name '*.html' -or \
+	               -name '*.txt' \
+	            \) -print | \
+	     xargs django-template-i18n-lint"
 
 test: build
 	$(API_ENDPOINT) docker-compose run django bash -c \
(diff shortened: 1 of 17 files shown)


SteveMarshall:
@marcofucci @matharden What d'you both think of this?

SteveMarshall:
Oh! Linting needs documenting in the `README`. I'll do that now…

SteveMarshall:
Retest this please

marcofucci:
Oh I'm sorry you had to do all of these, I didn't know you were actually changing the code.

A few things worth noticing:

 * lines of code are not 79 nowadays. In python, we tend to use the value 119 instead
 * there are a few error codes that are usually ignored as unnecessary and annoying

I usually tend to use the django coding style, see [here](https://docs.djangoproject.com/en/1.8/internals/contributing/writing-code/coding-style/) and their flake8 setup [here](https://github.com/django/django/blob/master/setup.cfg).

This means that you might have changed a bit of code for no good reason :-(

We can keep these changes if you want but I'd like to use the django flake8 `setup.cfg` instead.

matharden:
:+1:

SteveMarshall:
@marcofucci I've looked at the Django flake8 config, and the only things their rules ignore that I've fixed here are:

- Line lengths: I'd prefer leaving this at 79 and, if we find a really good reason to go longer later, do that.
- `E128 continuation line under-indented for visual indent`: This only appears to be an issue on the first parameter of `urlpatterns = patterns(…)` blocks. Given that [`patterns` is deprecated as of Django 1.8](https://docs.djangoproject.com/en/1.8/ref/urls/#patterns).

The rest of the changes I've made would've been required by the Django config, too, so I've not done *too* much wasted work. Given that, I'd err on the side of leaving this as-is (unless you and @matharden really object to 79 character lines?).

marcofucci:
@SteveMarshall that's because most of the changes are related to the line length.
We could talk about it for hours and still don't agree, I'm the only dev at the moment and I would very much like long lines but I guess you have made your decision so don't want to spend more time on this.

Re. the E128, I'd rather add it to the `setup.cfg` for the future and, if you want, fix the urlpatterns deprecation. I really don't want to keep the urlpatterns like that, it looks terrible.

I'll add the other Es and Ws to the ignore setting when we need in the future.

SteveMarshall:
@marcofucci You obviously feel strongly about this, so I'm happy to go with 119 columns and ignoring E128. I've unrolled those changes. You happy with the rest as they are?

marcofucci:
@SteveMarshall :+1: looks great. Are you happy / can I merge it?

SteveMarshall:
Ready to go! @matharden, in @marcofucci's absence, can you merge?

PR #33, opened 2015-07-28
Internal template

matharden:
This is mainly a better organisation of Sass files. I.e. each file is has single concern. 

It uses a new rebased version of template dependancies ([mojular](https://github.com/ministryofjustice/mojular/tree/internal) & [govuk-template](https://github.com/ministryofjustice/govuk-template)) and matches their directory structure.


SteveMarshall:
Looks good to me, @matharden, but “Better organised Mojular styles” is a bit of a clunky commit message :disappointed:

mtp_cashbook/assets-src/stylesheets/components/_subheader.scss
```diff
@@ -1,29 +0,0 @@
-.main-content {
-  padding-top: 0;
-}
-
-.global-subheader {
-  padding-top: 12px;
-  padding-bottom: 12px;
-  background: $grey-3-4;
```
teneightfive: Can we use a colour from the existing [govuk colour palette](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_colours.scss)?

mtp_cashbook/assets-src/stylesheets/elements/_subheader.scss
```diff
@@ -0,0 +1,9 @@
+.global-subheader {
+  padding-top: 12px;
+  padding-bottom: 12px;
+  background: $grey-3-4;
```
teneightfive: Can we use a colour from the existing [govuk colour palette](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_colours.scss)?

SteveMarshall:
Is this ready to be merged, @matharden?

PR #34, opened 2015-07-29
Alpha branding

teneightfive:
## User Story
As PM I'd like alpha branding on the site and tools so staff and citizens do not think it is a finished product ([#95958206](https://www.pivotaltracker.com/story/show/95958206))

**Note:** this merges into internal-template (#33).

PR #35, opened 2015-07-29
Notification style

teneightfive:
## User Story
As a monies clerk I'd like errors & flash messages to be clearly defined so I fix them appropriately - ([#93457302](https://www.pivotaltracker.com/story/show/93457302))

**Note:** this merges into alpha-branding branch (#34).

### DIFF
--- a/mtp_cashbook/templates/base.html
+++ b/mtp_cashbook/templates/base.html
@@ -41,10 +41,10 @@
   </p>
 
   {% if messages %}
-  <ul class="messages">
-      {% for message in messages %}
+  <ul class="messages" role="alert" tabindex="-1">
+    {% for message in messages %}
       <li{% if message.tags %} class="{{ message.tags }}"{% endif %}>{{ message }}</li>
-      {% endfor %}
+    {% endfor %}
   </ul>
   {% endif %}
 {% endblock %}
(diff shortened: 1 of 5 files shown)


mtp_cashbook/templates/base.html
```diff
@@ -32,11 +35,20 @@
   </div>
   {% endif %}
 
+  <div class="subheader-item phase-banner">
+    <span class="phase-tag phase-tag-alpha">
+      ALPHA
+    </span>
+    <span class="phase-banner-message">
+      This is a new service – your <a href="#">feedback</a> will help us to improve it.
```
SteveMarshall: Might be worth using `#TODO` instead of a raw `#` to make these easier to find? Also, maybe wrap all of `your feedback`?

SteveMarshall: Oh! And `blocktrans`… But I think this line is also in #34?

teneightfive: Will add the #TODO. Not sure about adding 'your' to the link though. The current [example](http://govuk-elements.herokuapp.com/#guide-alpha-beta) on the GOVUK element only links feedback.

SteveMarshall: :+1: Fair enough!

matharden:
LGTM :+1:

PR #40, opened 2015-08-12
Install en_US.UTF-8 locale to fix assets

mattmb:
See 6fcc62845dcbb43d3b90bcbf10721f6213da3716. This also installs the UTF
locale on the box and sets it in the LC_CTYPE env var as suggested here
https://github.com/sindresorhus/gulp-ruby-sass/issues/232


marcofucci:
:+1:

PR #41, opened 2015-08-13
Add running total to batch process view

teneightfive:
## User Story

As a monies clerk I'd like a running total of payments I've credited so I don't have to calculate it myself - ([98802420](https://www.pivotaltracker.com/story/show/98802420))

### DIFF
--- a/mtp_cashbook/assets-src/javascripts/modules/moj.running-total.js
+++ b/mtp_cashbook/assets-src/javascripts/modules/moj.running-total.js
@@ -1,0 +1,67 @@
+// Running Total module
+// Dependencies: moj, _, jQuery
+
+(function () {
+  'use strict';
+
+  moj.Modules.RunningTotal = {
+    selector: '.js-RunningTotal',
+
+    init: function () {
+      _.bindAll(this, 'render', 'updateTotal');
+      this.cacheEls();
+
+      if (this.$totalsContainer.length) {
+        this.bindEvents();
+        this.render();
+        this.updateTotal();
+      }
+    },
+
+    cacheEls: function () {
+      this.$items = $('.js-RunningTotal-item');
+      this.$totalsContainer = $(this.selector);
+      this.label = this.$totalsContainer.data('label');
+
+      this.$label = $('<p>');
+      this.$total = $('<strong>');
+    },
+
+    bindEvents: function () {
+      moj.Events.on('RunningTotal.render', this.render);
+      this.$items.on('change', this.updateTotal);
+    },
+
+    updateTotal: function () {
+      this.$total.html('&pound;' + this.calculateTotal());
+      this.$label
+        .text(this.label)
+        .append(this.$total);
+    },
+
+    calculateTotal: function () {
+      var total = 0;
+
+      this.$items.each(function (i, el) {
+        var $el = $(el);
+        var amount = $el.data('amount');
+
+        if ($el.is(':checked')) {
+          total += parseFloat(amount);
+        }
+      });
+
+      return total.toFixed(2);
+    },
+
+    render: function () {
+      this.$label
+        .attr('aria-live', 'polite')
+        .attr('aria-atomic', 'false');
+
+      this.$totalsContainer
+        .append(this.$label);
+    }
+  };
+
+})();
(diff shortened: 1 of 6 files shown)


mtp_cashbook/templates/cashbook/transaction_batch_list.html
```diff
@@ -43,7 +43,7 @@
           <td class="numeric">&pound;{{ transaction.amount|currency }}</td>
           <td>{{ transaction.sender }}</td>
           <td class="check">
-            <input type="checkbox" name="{{ form.transactions.html_name }}" value="{{ transaction_pk }}" id="check-{{ transaction.id }}" {% if transaction_pk|to_string in form.transactions.value %}checked{% endif %}>
+            <input type="checkbox" name="{{ form.transactions.html_name }}" value="{{ transaction_pk }}" id="check-{{ transaction.id }}" {% if transaction_pk|to_string in form.transactions.value %}checked{% endif %} class="js-RunningTotal-item" data-amount="{{ transaction.amount|currency }}">
```
SteveMarshall: Can we break this line up a bit, @teneightfive? Maybe split between attributes or something?

mtp_cashbook/assets-src/javascripts/modules/moj.running-total.js
```diff
@@ -0,0 +1,57 @@
...
+        if ($el.is(':checked')) {
+          total += parseFloat(amount);
+        }
+      });
+
+      return total.toFixed(2);
+    },
+
+    render: function () {
+      this.$totalsContainer.append('<p>Total payments you have credited: <strong class="js-RunningTotal-amount">&pound;0.00</strong></p>');
```
SteveMarshall: Can we get the markup/text in this passed in from config, and get the text translated?

mtp_cashbook/assets-src/stylesheets/views/_batch.scss
```diff
@@ -3,12 +3,17 @@
   margin-bottom: 1em;
 
   p {
-    display: inline;
     font-size: 24px;
+    margin-top: 0;
+
+    @include breakpoint(769px) {
```
SteveMarshall: Why `769px`? Is there some way we can make it clearer what this breakpoint is for?

teneightfive: 769 is the dimension GDS use as desktop width. This is using @tyom's new mojular and I don't think it has a way of clearly showing it's a desktop width.

I can look at moving these values into something a bit clearer.

teneightfive:
@SteveMarshall what are your thoughts on the changes?

I've moved the label to the template and wrapped it in `trans`, not sure I'm 100% happy with it being a `data` attribute. Any other suggestions?

I've also improved the accessibility of the total so that Assistive Technologies will read out the total again when it gets updated.

SteveMarshall:
The breakpoint clarification looks good! What about doing the `trans` bits like [elsewhere in the tour](https://github.com/ministryofjustice/money-to-prisoners-cashbook/blob/master/mtp_cashbook/templates/cashbook/dashboard.html#L41)? Create a JS variable in an include somewhere in the template, put it into a content block, and `trans` it?

teneightfive:
I think I prefer having the text as an attribute to setting a local JS variable in the view. Keeps it joined together with the class which is executing the module.

SteveMarshall:
Fair enough; I can certainly live with this. My only suggestion might be to include the substitution marker for the value in the attribute, like this: `{% trans 'Total payments you have credited: { highlight }{ total }{end highlight}' %}` (or similar), in case, when changing the translation, we decide we need to change the position of the total in the string.

teneightfive:
I would argue that we should approach that if the need arises later on. It feels like we're going to want to keep the figure after the actual label even in other language. 

I don't think there's a need to over complicate the code at this stage based on the chance of something happening. If we need to change it at a later stage then doing it then will be clearer why it was changed.

SteveMarshall:
@teneightfive Fair enough; let's leave as-is then. You wanna squash and push?

teneightfive:
@SteveMarshall good to merge when you're ready

PR #42, opened 2015-08-14
Batch print layout

teneightfive:
## User Story

As a monies clerk I'd like the batch view to be printable, because I have to print - ([#99023134](https://www.pivotaltracker.com/story/show/99023134))

### DIFF
--- a/mtp_cashbook/assets-src/javascripts/modules/moj.dialog.js
+++ b/mtp_cashbook/assets-src/javascripts/modules/moj.dialog.js
@@ -1,0 +1,101 @@
+// Dialog module
+// Dependencies: moj, _, jQuery
+
+/* globals console */
+(function () {
+  'use strict';
+
+  moj.Modules.Dialog = {
+    selector: '.js-Dialog',
+
+    init: function () {
+      _.bindAll(this, 'render', 'closeDialog', 'onKeyUp');
+      this.cacheEls();
+      this.bindEvents();
+    },
+
+    cacheEls: function () {
+      this.$body = $('body');
+      this.$backdrop = $('<div>').addClass('Dialog-backdrop');
+      this.$close = $('<a>');
+
+      // close element
+      this.$close
+        .attr('href', '#')
+        .addClass('Dialog-close');
+    },
+
+    bindEvents: function () {
+      moj.Events.on('Dialog.render', this.render);
+      this.$body.on('click', this.selector, this.render);
+    },
+
+    render: function (e) {
+      this.$triggerEl = $(e.target);
+
+      e.preventDefault();
+      this.openDialog(this.$triggerEl.attr('href'));
+    },
+
+    openDialog: function (target) {
+      var $dialog = $(target);
+
+      // log warning if target doesn't exist
+      if ($dialog.length === 0) {
+        moj.log('Modules.Dialog Warning: The target element, ' + target + ', does not exist.');
+        return;
+      }
+
+      if ($dialog.data('close-label')) {
+        this.$close.text($dialog.data('close-label'));
+      } else {
+        this.$close.text('close');
+      }
+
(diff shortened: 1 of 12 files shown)


mtp_cashbook/templates/cashbook/transaction_batch_list.html
```diff
@@ -65,6 +74,22 @@
...
+      <p>You could also try using the ALT + TAB keys to switch between the windows.</p>
+      {% endblocktrans %}
+      <button class="button u-nojs-hidden js-Print">{% trans "Print now" %}</button>
+
+      <p>
+        <input type="checkbox" name="remove-print-prompt" id="remove-print-prompt" value="1">
+        <label for="remove-print-prompt">{% trans "Don't show this message again" %}</label>
+      </p>
+    </div>
+  </dialog>
```
SteveMarshall: It might be nice to add these bits of JS-only print UI with javascript, rather than in HTML.

mtp_cashbook/templates/cashbook/transaction_batch_list.html
```diff
@@ -1,6 +1,8 @@
 {% extends "base.html" %}
 {% load currency core i18n %}
 
+{% block template_type %}v-batch{% endblock %}
```
SteveMarshall: `v` for… view?

mtp_cashbook/assets-src/javascripts/modules/moj.dialog.js
```diff
@@ -0,0 +1,107 @@
...
+      e.preventDefault();
+      this.openDialog(this.$triggerEl.attr('href'));
+    },
+
+    openDialog: function (target) {
+      var $dialog = $(target);
+
+      // log warning if target doesn't exist
+      if ($dialog.length === 0) {
+        console.warn('The target element, ' + target + ', does not exist.');
```
SteveMarshall: Is this gonna break/error in IE8?

mtp_cashbook/templates/cashbook/transaction_batch_list.html
teneightfive: I built it so that they're not JS only. If you turn JS off you won't get the button for the dialog but the dialog will appear in an attempt to deter you from printing unnecessarily. Have a look without JS and see what you think.

mtp_cashbook/templates/cashbook/transaction_batch_list.html
teneightfive: Yep

mtp_cashbook/assets-src/javascripts/modules/moj.dialog.js
teneightfive: Good point. I was trying to give informative errors if a target doesn't exist. Might just remove this block?

mtp_cashbook/templates/cashbook/transaction_batch_list.html
SteveMarshall: Is this shorthand used much elsewhere? Is there a reason we can't use `view-batch` (or even `batch-view`)?

mtp_cashbook/assets-src/javascripts/modules/moj.dialog.js
SteveMarshall: If you find it useful, keep it… but maybe appropriately wrap it in conditionals?

mtp_cashbook/templates/cashbook/transaction_batch_list.html
teneightfive: Yes, I've used it previously and it's similar to conventions like using `u-` for a [utility](https://github.com/ministryofjustice/govuk-template/blob/master/assets/scss/govuk/_utilities.scss).

PR #45, opened 2015-08-18
Batch print mode improvements

teneightfive:


SteveMarshall:
What's going on with the message on the first commit here? “Fix!”? Other than that, looks good to me.

PR #50, opened 2015-08-24
Sticky totals

teneightfive:
## User story [#99020392](https://www.pivotaltracker.com/story/show/99020392)

As a monies clerk I'd like the control total to remain visible at all times so it can be monitored

### DIFF
--- a/mtp_cashbook/assets-src/javascripts/modules/moj.sticky-header.js
+++ b/mtp_cashbook/assets-src/javascripts/modules/moj.sticky-header.js
@@ -1,0 +1,68 @@
+// Sticky header module
+// Dependencies: moj, _, jQuery
+
+(function () {
+  'use strict';
+
+  moj.Modules.StickyHeader = {
+    selector: '.js-StickyHeader',
+
+    stickyClass: 'is-sticky',
+
+    init: function () {
+      _.bindAll(this, 'render', 'onScroll');
+      this.cacheEls();
+
+      if (this.$originalHeader.length) {
+        this.bindEvents();
+        this.render();
+      }
+    },
+
+    cacheEls: function () {
+      this.$originalHeader = $(this.selector);
+
+      if (this.$originalHeader.length) {
+        this.$window = $(window);
+        this.$body = $('body');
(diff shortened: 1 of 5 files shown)


mtp_cashbook/assets-src/javascripts/modules/moj.sticky-header.js
```diff
@@ -0,0 +1,68 @@
...
+        if (!this.$stickyHeader.is(':visible')) {
+          this.$stickyHeader
+            .css({
+              top: -this.$originalHeader.height()
+            })
+            .show()
+            .stop()
+            .animate({
+              top: 0
+            }, 500);
```
SteveMarshall: I assume the `500` here is half a second? Maybe worth pulling out into a variable?

mtp_cashbook/assets-src/javascripts/modules/moj.sticky-header.js
```diff
@@ -0,0 +1,68 @@
...
+        }
+      } else {
+        this.$stickyHeader
+          .stop()
+          .animate({
+            top: -this.$originalHeader.height()
+          }, 150, 'linear', function () {
```
SteveMarshall: What's the `150` here?

mtp_cashbook/assets-src/javascripts/modules/moj.sticky-header.js
teneightfive: Yep, it's the animation duration. Not sure it's worth making that into another variable. The docs/examples for `.animate()` are clear enough to know what this is.

mtp_cashbook/assets-src/javascripts/modules/moj.sticky-header.js
teneightfive: Same as above, it's the duration for the animation.

SteveMarshall: :+1:

SteveMarshall:
Looks good to me, @teneightfive. @tonyduarte, you happy with it?

PR #51, opened 2015-08-24
Add batch actions (print/select all) top and bottom

teneightfive:
This follows up on a conversation had around "[As a monies clerk I'd like the batch view to be printable, because I have to print](https://www.pivotaltracker.com/story/show/99023134)" to include the print button at the top and bottom of the transaction list.


mtp_cashbook/templates/cashbook/includes/table_actions.html
```diff
@@ -0,0 +1,10 @@
+{% load i18n %}
+<div class="batch-actions">
+  <p class="print">
+    <a href="#print-dialog" class="js-Dialog">{% trans "Print these transactions" %}</a>
+  </p>
+  <p class="select-all">
+    <input type="checkbox" id="select-all-{{ location }}" class="Checkbox--right js-SelectAll" data-name="{{ field_name }}">
+    <label for="select-all-{{ location }}">{% trans "Select all" %}</label>
+  </p>
+</div>
```
SteveMarshall: Can we add this via JS, as it doesn't work without?

mtp_cashbook/templates/cashbook/transaction_batch_list.html
```diff
@@ -55,19 +57,9 @@
...
-          </th>
-          <th colspan="2" class="select-all">
-            <input type="checkbox" id="select-all" class="Checkbox--right">
-            <label for="select-all">{% trans "Select all" %}</label>
-          </th>
-        </tr>
-      </tfoot>
     </table>
+
+    {% include "cashbook/includes/table_actions.html" with location='footer' field_name=form.transactions.html_name %}
```
SteveMarshall: Why not put these in `thead` and `tfoot`? Strikes me that the select all, at least, would make more sense (and work better for users of assistive tech?) in those…

teneightfive: Yep, they are linked to the contents of the table so would make more sense within `thead` and `tfoot`.

mtp_cashbook/templates/cashbook/includes/table_actions.html
teneightfive: As discussed yesterday, I'm not about this right now. 

Without JS they are currently hidden from the view and moving them to JS would either mean having the markup in JS files or adding JS templating, which might increase the complexity around internationalization. 

I'm happy either way but at the moment leaning towards what we have and not involving JS templates as it keeps all the i18n handled by django. 

What's your preference?

SteveMarshall: Leaving as-is works for me, just so long as we've considered our reasons for one choice or the other!

teneightfive:
fixup! Moved the batch actions into `thead` and `tfoot` and refactored some of the sass to compensate.

teneightfive:
@SteveMarshall okay to rebase and then merge?

PR #52, opened 2015-08-27
Feature tour enhancements

teneightfive:
- Remove JS error when no `tour` object exists
- Update hopscotch to avoid multiple highlight containers being added even when one already exists or the tour isn't active.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,6 +15,6 @@
     "hopscotch": "~0.2.5",
     "js-cookie": "~2.0.3",
     "jquery": "~1.11",
-    "hopscotch-highlight": "ministryofjustice/hopscotch-highlight#0.1.0"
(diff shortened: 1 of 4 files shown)


SteveMarshall:
This looks good, but there are still a bunch of situations where the tour popups get hidden. Notably, the prisoner number column's tour tip on a browser window at 800×600, but there are others too. Do we care for this PR?

teneightfive:
Yeah, I'm not 100% happy with how the popups are handled on smaller resolutions. The current style for the box assumes it should come off the middle of the element which means some, like the prisoner number, can get cut off the screen slightly. 

Probably should have a look in this branch.

SteveMarshall:
LGTM now :+1:

PR #57, opened 2015-09-11
Unload prompt

teneightfive:
# User story [#103089538](https://www.pivotaltracker.com/story/show/103089538)

As a monies clerk I want to be warned if I leave the batch screen and I have ticked some transactions so I submit my credited transactions.

[finishes #103089538]

### DIFF
--- a/mtp_cashbook/templates/cashbook/transaction_batch_list.html
+++ b/mtp_cashbook/templates/cashbook/transaction_batch_list.html
@@ -4,7 +4,11 @@
 {% block template_type %}v-batch{% endblock %}
 
 {% block content %}
-  <form method="post" action="">
+  <form
+    method="post"
+    action=""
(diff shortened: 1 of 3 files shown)


mtp_cashbook/templates/cashbook/transaction_batch_list.html
```diff
@@ -4,7 +4,7 @@
 {% block template_type %}v-batch{% endblock %}
 
 {% block content %}
-  <form method="post" action="">
+  <form method="post" action="" class="js-BeforeUnload" data-unload-msg="{% trans 'You are about to lose your ticked payments. If you have credited NOMIS with the selected prisoner payments, you must submit them.' %}">
```
SteveMarshall: Can we break this over multiple lines?

PR #61, opened 2015-09-21
Batch processing validation

teneightfive:
## User Story [#103513830](https://www.pivotaltracker.com/story/show/103513830)
As UR I would like the Digital Cash Book updated so that it is ready for user research

This PR adds some validation features around batch processing to closer match original prototype.


SteveMarshall:
:bomb:…:boom:!

PR #62, opened 2015-09-21
Locked transactions view only shows the ones locked by user

marcofucci:
This fixes a bug caused by the cashbook client not filtering by
logged-in user when requesting locked transactions.
This caused prison clerks from the same prison to see all the locked
transactions.


marcofucci:
@ibrechin can you take a look?

PR #63, opened 2015-09-23
Functional/Integration tests

teneightfive:
## User Story [#103485730](https://www.pivotaltracker.com/story/show/103485730)
As a developer,
I want the cashbook to be able to run integration/functional tests,
So that I know I haven't broken anything with my changes


test/features/pages/signin.js
```diff
@@ -0,0 +1,58 @@
...
+      .setValue(this.passwordField, password)
+      .submitForm(this.signinForm)
+      .call(callback);
+  };
+
+  this.signin = function (callback) {
+    var that = this;
+
+    this.get(function () {
+      that.fillForm(testUsername, testPassword, callback);
```
SteveMarshall: Is there a reason not to have this just call `this.signinWith(testUsername, testPassword, callback)`?

test/features/pages/signin.js
```diff
@@ -0,0 +1,58 @@
...
+
+  this.signinWith = function (username, password, callback) {
+    this.fillForm(username, password, callback);
+  };
+
+  this.signout = function (callback) {
+    browser
+      .url('/auth/logout/')
+      .call(callback);
+  };
```
SteveMarshall: I wonder if just changing the URL to `/auth/logout/` is a true test of signing out? Might it be better to have this click the link on the page?

test/features/steps/shared.js
```diff
@@ -0,0 +1,81 @@
...
+   *
+   * Usage: Then I should see "some text"
+   *
+   * @params {string} text The text to check for
+   * @params {function} next The callback function of the scenario
+   */
+  this.Then(/^I should see "([^"]*)"$/, function(text, next) {
+    browser
+      .getText('body')
+      .should.eventually.contain(text)
```
SteveMarshall: `eventually` :smile:

test/features/support/hooks.js
```diff
@@ -0,0 +1,38 @@
...
+   * BeforeFeatures handler
+   *
+   * Provides access to the pageObjectsMap
+   * on browser
+   *
+   * @params {event} event The event being attached to
+   * @params {function} callback The callback function of the hook
+   */
+  this.registerHandler('BeforeFeatures', function (event, callback) {
+    browser.pageObjects = require('./pageObjectMap.js');
```
SteveMarshall: Might this be better called simply `pages`?

test/wdio.conf.js
```diff
@@ -0,0 +1,114 @@
...
+  }, {
+    browserName: 'firefox'
+  }, {
+    browserName: 'phantomjs'
+  }],
+
+  // ===================
+  // Test Configurations
+  // ===================
+  logLevel: 'silent',
```
SteveMarshall: Do we maybe want more logging than this?

test/wdio.conf.js
```diff
@@ -0,0 +1,114 @@
...
+  screenshotPath: screenshotPath,
+  baseUrl: 'http://localhost:8001',
```
SteveMarshall: Is this the URL you're testing? Is it worth this being configurable at runtime?

SteveMarshall:
This is really good, @teneightfive! :raised_hands: 

Is it worth adding a test for something like:

```
As a user
I need to sign in before I can access the dashboard
So that I can't see anything I shouldn't
```

test/features/pages/signin.js
teneightfive: Seems unnecessary as `siginWith` is only calling `fillForm` anyway? This skips having to call another method.

test/features/pages/signin.js
teneightfive: I was planning to change to that, but annoyingly, because of the feature tour, in some cases you'd have to click dismiss first and then the sign out link.

Need to find a better way to do that process. Possibly login with a cookie set that will disable the feature tour for the purpose of this test?

test/features/support/hooks.js
teneightfive: Did consider that, but they're normally referred to as [Page Objects](http://blog.josephwilk.net/cucumber/page-object-pattern.html) when using this pattern so that made more sense.

test/wdio.conf.js
teneightfive: Not sure we do at this stage. This is the level of output that is displayed for the selenium interactions and are actually quite noisy around the basic test logging.

See [logLevel](http://webdriver.io/guide/getstarted/configuration.html#logLevel)

test/wdio.conf.js
teneightfive: Yep! Very good spot. Will sort that so it can be an ENV or fallback to this.

teneightfive:
@SteveMarshall sounds like it would be the same as the `user_sigin` [test](https://github.com/ministryofjustice/money-to-prisoners-cashbook/pull/63/files#diff-1d036b33ecc5a2a9f2e9e72c1d63c0eaR2).

Maybe a case of just re-wording that if you're not happy with that wording?

test/features/pages/signin.js
SteveMarshall: Yeah, that sounds about right… Though we might want to be explicit about that in the tests? (Like `with the feature tour disabled` or something?)

test/features/support/hooks.js
SteveMarshall: If you look at the code in that page, though, it doesn't refer to “page objects”, it just calls them `something_page` or `SomethingPage`. You wouldn't call an object that follows the [singleton pattern](https://en.wikipedia.org/wiki/Singleton_pattern) `SomethingSingleton` (unless you were writing Java, I guess), so I think `pageObject` is wrong: you're leaking the name of the pattern into your code.

test/wdio.conf.js
SteveMarshall: Ah, right: I wasn't sure if this was for Selenium or for your tests. In that case, yeah, silent is fine. :+1:

test/features/pages/signin.js
SteveMarshall: Hmm. I wonder if actually it'd be nicer as one function with default values for username and password? That way, the intent is much clearer and you're not spreading “sign in” functionality across two functions.

SteveMarshall:
I think it's a different scenario, actually, or maybe in the tests for the dashboard (and other pages).

Something like:

```
Scenario: Haven't signed in, so can't do stuff
  Given I am not signed in
  When I go to the "Dashboard" page
  Then I should see "You are not logged in"
```

test/features/pages/signin.js
teneightfive: But then you would end up with having to call the method with null values if you wanted to use the default creds: `signin(null, null, callback)`. 

Unless you flip the callback param to the front but that breaks the pattern being used elsewhere and by cucumber.

test/features/pages/signin.js
teneightfive: Yep, can add a `Given` statement for that.

test/features/pages/signin.js
SteveMarshall: Oh, god, yeah, of course it does. Roll on ES6, I guess? Leave it as-is, I suppose!

test/features/support/hooks.js
teneightfive: If we renamed this to `pages` would you then suggest renaming the file to `pageMap.js` and removing any reference to `object` in that file?

teneightfive:
Yeah, that could be a different scenario in that case. 

We need a separate story for writing out all the current scenarios too.

test/features/pages/signin.js
teneightfive: Yep, would be much nicer with ES6. Was actually considering using coffee for some of this as it is a bit nicer for writing tests. 

But assumed it might be better to not add too many more layers at this point.

test/features/support/hooks.js
SteveMarshall: Yeah, I reckon so :+1:

test/features/pages/signin.js
SteveMarshall: Yeah, let's leave it as-is; less layers == better.

teneightfive:
retest this please

teneightfive:
@SteveMarshall just had a thought about this...shall we get this working with Jenkins before we merge?

SteveMarshall:
@teneightfive I think the way to get it “working with Jenkins” is going to be to get it working with a pure docker-compose setup, then have a `make integration-tests` type thing?

test/features/pages/signin.js
maxf: it should be `/logout`, not `/auth/logout`, no?
Ditto for `auth/login` in signin.js

test/features/user_signout.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Signing out
+  As a user
+  I want to be able to sign out
+  So that I do not leave access to my account open
+
+  Scenario: Successful sign out
+    Given I am signed in
+    When I sign out
+    Then I should see "Digital cashbook" as the page title
```
maxf: cashbook was capitalised since

test/features/pages/signin.js
```diff
@@ -0,0 +1,58 @@
...
+var testPassword = 'test_prison_1';
+
+var SigninPage = function () {
+  this.usernameField = '#id_username';
+  this.passwordField = '#id_password';
+  this.signinForm = '#login-form';
+
+  this.get = function (callback) {
+    browser
+      .url('/auth/login/')
```
maxf: now /login

SteveMarshall:
@teneightfive @maxf One of you wanna fix the remaining bits on here?

maxf:
@SteveMarshall fixed some remaining bits. Note that one test currently fails because of #104729954

bower.json
```diff
@@ -10,6 +10,10 @@
     "tests"
   ],
   "dependencies": {
+<<<<<<< HEAD
+    "money-to-prisoners-common": "ministryofjustice/money-to-prisoners-common#0.2.2"
+=======
     "money-to-prisoners-common": "ministryofjustice/money-to-prisoners-common#0.3.0"
+>>>>>>> origin
```
SteveMarshall: I'm not sure you wanted to commit this, @maxf?

maxf:
retest this please

SteveMarshall:
LGTM; rebase and merge?

teneightfive:
:+1:

PR #67, opened 2015-09-28
Remove auth prefix from login urls

ibrechin:
For consistency with the other front-end applications.


ibrechin:
@SteveMarshall beep boop

PR #70, opened 2015-10-01
only show back button when transaction list empty

maxf:


mtp_cashbook/templates/cashbook/includes/batch_buttons.html
```diff
@@ -1,5 +1,7 @@
 {% load i18n %}
 <div>
-  <button type="submit" name="submit" value="submit" class="button" id="tour-step6">{% trans "I've credited ticked payments in NOMIS" %}</button>
+  {% if object_list %}
+    <button type="submit" name="submit" value="submit" class="button" id="tour-step6">{% trans "I've credited ticked payments in NOMIS" %}</button>
+  {% endif %}
```
SteveMarshall: Do we need the `if` here? You're already including this in `transaction_batch_list` inside an `if object_list`…

maxf: The alternative is to have two partials: one with a single button (for when the list is empty) and one with both buttons. 
But then the back button is duplicated across both partials.
You could always DRY that by having a new partial just for the submit button, that's optionally called by the back_buttons partial, but that's arguably overkill.
That's why Dom and I agreed to do it this way.

SteveMarshall: Ahhhh, right! Fine by me, then!

PR #73, opened 2015-10-05
Change empty batch submission validation to server side

teneightfive:
## User Story [#104079856](https://www.pivotaltracker.com/story/show/104079856)

As a monies clark,
I want to be given an error when I submit a batch without ticking any payment,
So that I know I have made an error

[Finishes #104079856]

### DIFF
--- a/mtp_cashbook/templates/cashbook/transaction_batch_list.html
+++ b/mtp_cashbook/templates/cashbook/transaction_batch_list.html
@@ -12,62 +12,83 @@
     data-unload-msg="{% trans "You are leaving this page, but have ticked payments. &#10;&#10;You must click &#8216;I&#8217;ve credited payments in NOMIS&#8217; if payments have been credited." %}">
     {% csrf_token %}
 
-    <div class="totals js-StickyHeader"  id="tour-step-total">
-      <div class="container js-RunningTotal" data-label="{% trans 'Total payments you have credited: ' %}">
-        <p>
-          {% blocktrans with total=total|currency highlight="<strong>&pound;" highlightend="</strong>" %}
-            Control total: {{ highlight }}{{ total }}{{ highlightend }}
-          {% endblocktrans %}
-        </p>
-      </div>
-    </div>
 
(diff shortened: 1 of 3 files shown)


mtp_cashbook/templates/cashbook/transaction_batch_list.html
```diff
@@ -12,62 +12,83 @@
...
-          {% endblocktrans %}
-        </p>
-      </div>
-    </div>
 
     {% if object_list %}
-    {{ form.errors }}
-
-    {% include "cashbook/includes/batch_buttons.html" %}
+      {% if form.errors or form.non_field_errors %}
```
teneightfive: This block could be extracted out into an include and added to [money-to-prisoners-common](http://github.com/ministryofjustice/money-to-prisoners-common) if it can be used across all 3 apps.

A slight variation is already being used in the [login template](https://github.com/ministryofjustice/money-to-prisoners-common/blob/4343e219e86acfa55ef0ac2ea630e767f94a8d48/templates/auth/login.html#L14-L35).

SteveMarshall: I'd be wary of extracting it too soon: it might be that it's not *actually* common between them, it just looks it now, if you take my meaning?

teneightfive: Yep, I wasn't suggesting doing it right now, more a case of making a note against this change to make @maxf aware that it might need to be done in the future. 

*(Probably looked like I was just critiquing my own work)*

teneightfive:
retest this please

mtp_cashbook/templates/cashbook/transaction_batch_list.html
SteveMarshall: Yeah, it looked exactly like you were critiquing your own work… Which is a perfectly good and valid thing to do: I do it all the time… When I get to do any coding, that is :disappointed:

PR #74, opened 2015-10-06
avoid installing unnecessary packages

maxf:
Don't install browser-sync and other dependent packages
on non-local environments, as those packages are only
useful on developer machines
[updates #104986742]


maxf:
retest this please

maxf:
retest this please

PR #76, opened 2015-10-06
Add jenkins config files for running tests

marcofucci:
This configures django-jenkins so that jenkins can run unit tests and generate reports.


marcofucci:
@ibrechin do you mind?

PR #84, opened 2015-10-27
make gulp serve work in docker container

maxf:
In order to detect changes in money-to-prisoners-common, we mount it in /opt
(this requires the repo to be next to money-to-prisoners-cashbook). Then we
tell gulp to track changes in /opt and tell bower where to look for assets
to import using bower-link. Node-sass needs to be rebuilt, for some reason.
The delay before syncing browsers must also be increased to 3s, as it takes
longer in the container to gulp all the things. (The delay isn't a good idea
in the first place. We'll need to fix that.)

Maybe there's a better way to do this. Let's discuss.

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,3 +1,4 @@
 {
-  "directory": "mtp_cashbook/assets-src/bower_components"
+  "cwd": "/usr/lib/node_modules/money-to-prisoners-cashbook",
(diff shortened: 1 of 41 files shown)


SteveMarshall:
@maxf Given your work to remove Bower and reduce our front-end debt, can we kill this and re-visit later?

PR #85, opened 2015-10-29
update home page copy [closes #106947908]

maxf:
From changes on https://docs.google.com/document/d/1uAj9CbCS3MMUSVLaMLtqT4qrnYtjZa0wb_nGLxaB3cI/edit

### DIFF
--- a/mtp_cashbook/templates/cashbook/dashboard.html
+++ b/mtp_cashbook/templates/cashbook/dashboard.html
@@ -10,7 +10,7 @@
         <h2>{% trans "New" %}</h2>
         <p>
           {% blocktrans with payments_count="<strong>{0}</strong>"|safewrap:new_transactions %}
-          {{ payments_count }} Payments to credit
+          {{ payments_count }} Credits to process
           {% endblocktrans %}
         </p>
       </a>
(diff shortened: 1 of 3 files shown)


mtp_cashbook/templates/cashbook/dashboard.html
```diff
@@ -10,7 +10,7 @@
         <h2>{% trans "New" %}</h2>
         <p>
           {% blocktrans with payments_count="<strong>{0}</strong>"|safewrap:new_transactions %}
-          {{ payments_count }} Payments to credit
+          {{ payments_count }} Credits to process
```
SteveMarshall: It might be nice if we could do the capitalisation here in CSS (`Credits` is the second word in the sentence), but I can't see a clean way to do that without extra markup?

mtp_cashbook/templates/cashbook/dashboard.html
```diff
@@ -19,20 +19,16 @@
...
     </li>
 
     <li class="SummaryBoxes--locked">
       <a href="{% url 'transactions-locked' %}">
-        <h2>{% trans "Locked" %}</h2>
+        <h2>{% trans "In progress" %}</h2>
         <p>
           {% blocktrans with payments_count="<strong>{0}</strong>"|safewrap:locked_transactions %}
-          {{ payments_count }} Payments currently locked
+          {{ payments_count }} Credits in progress
```
SteveMarshall: Same here as for `Credits to process` above.

mtp_cashbook/templates/cashbook/dashboard.html
maxf: That's right, you can capitalise every word in a span, not just the first.

PR #91, opened 2015-11-03
Fix user login cucumber integration test

ushkarev:
generated users have a different naming scheme now
depends on [api#57](https://github.com/ministryofjustice/money-to-prisoners-api/pull/57)


maxf:
Does that mean that that test isn't currently passing in master?

PR #94, opened 2015-11-04
Ignore correctly the directory where static assets are collected

ushkarev:


ibrechin:
Should this also apply to the .dockerignore file?

ushkarev:
it doesn't actually matter. `RUN ./manage.py collectstatic --noinput` runs after `ADD . /app` so there won't be any static files yet. but for consistency, i'll add it :)

PR #101, opened 2015-11-13
Use newer MTP gulp tasks to fix CSS minification errors

ushkarev:


package.json
```diff
@@ -26,8 +27,8 @@
     "bower": "latest",
     "gulp": "^3.9.0",
     "main-bower-files": "^2.9.0",
-    "nconf": "^0.7.2",
-    "money-to-prisoners-gulp-tasks": "ministryofjustice/money-to-prisoners-gulp-tasks#0.1.0"
+    "money-to-prisoners-gulp-tasks": "ministryofjustice/money-to-prisoners-gulp-tasks#0.2.0",
```
maxf: I think money-to-prisoners-gulp-tasks should currently be at 0.0.2 since the last change was a bug fix, not a new feature.

ushkarev: do you mean 0.1.1? i still kinda think it's at least a minor version change rather than a patch level change… it's not really backwards compatible even…

PR #107, opened 2015-11-23
Make transaction history search form work via a GET request

ushkarev:
so that google analytics can be used to monitor search parameters such, particularly the text-based search

### DIFF
--- a/mtp_cashbook/apps/cashbook/forms.py
+++ b/mtp_cashbook/apps/cashbook/forms.py
@@ -116,10 +116,10 @@ class DiscardLockedTransactionsForm(forms.Form):
 
 
 class FilterTransactionHistoryForm(forms.Form):
-    received_at_0 = forms.DateField(required=True, label=_('Received at start date'),
-                                    widget=MtpDateInput)
-    received_at_1 = forms.DateField(required=True, label=_('Received at end date'),
-                                    widget=MtpDateInput)
+    start = forms.DateField(required=True, label=_('Received at start date'),
+                            widget=MtpDateInput)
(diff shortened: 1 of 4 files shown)


maxf:
Can I suggest to make the URL nicer?
instead of :
?received_at_0=16%2F11%2F2015&received_at_1=23%2F11%2F2015&search=blah&owner=
better:
?from=20151116&to=20151120&search=blah&owner=me

maxf:
Can you confirm that django prevents DOM/script injections from the query string params?

ushkarev:
RE injection: django escapes all form values unless explicitly marked as safe

ushkarev:
RE GET params: we can change `received_at_0` and `received_at_1` but the number following that depends on what's input into the form field. it'd be weird to make people type dates specifically as `20151116`… unless you set up a date picker and then pre-process the form on submission with JS

maxf:
We can use tricks not to show the user confusing date formats on the form
while having them in the URL, but let's not bother just yet.

SteveMarshall:
In terms of the query string param formats, @ushkarev is right: we need a sensible format in the front-end. Ideally, something hyphen-separated (slashes get escaped in query strings).

I might also argue for `from` and `to` as the parameter names, rather than `since` and `until`, because they seem clearer? Or is there something I'm missing?

ushkarev:
max and i were thinking that the date formatting is a slightly separate
issue which could go with the JS date picker task. the primary point of
this PR was to put the search text into the url for google analytics to
pick up.

ushkarev:
as we were saying, `from` is a reserved keyword so it'd be overly hacky to
make it work.

PR #110, opened 2015-11-26
Use the latest history page designs and copy

maxf:
- make the search box look as intended
- readd the show-hide code and make it work in IE8
- fix the search box markup to make changes easier
- remove unnecessary table columns
- add tests

### DIFF
--- a/mtp_cashbook/assets-src/javascripts/main.js
+++ b/mtp_cashbook/assets-src/javascripts/main.js
@@ -1,4 +1,4 @@
-/* globals require, console */
+/* globals require */
 
 (function() {
   'use strict';
--- a/mtp_cashbook/assets-src/javascripts/main.js
+++ b/mtp_cashbook/assets-src/javascripts/main.js
@@ -11,6 +11,7 @@
       require('mojular-moj-elements'),
       require('dialog'),
       require('batch-validation'),
+      require('show-hide'),
       require('feature-tour'),
       require('messages'),
       require('print'),
@@ -22,11 +23,11 @@
     ])
     .init();
 
-  var moduleNames = Object.keys(Mojular.Modules);
-  if(moduleNames.length) {
-    var moduleList = moduleNames.map(function(i) {
-      return '• ' + i;
-    }).join('\n').replace(/^\s+/, '');
-    console.log('The following modules are loaded:\n' + moduleList);
-  }
+  // var moduleNames = Object.keys(Mojular.Modules);
+  // if(moduleNames.length) {
+  //   var moduleList = moduleNames.map(function(i) {
+  //     return '• ' + i;
+  //   }).join('\n').replace(/^\s+/, '');
+  //   console.log('The following modules are loaded:\n' + moduleList);
+  // }
 }());
--- a/mtp_cashbook/assets-src/javascripts/main.js
+++ b/mtp_cashbook/assets-src/javascripts/main.js
@@ -22,12 +22,4 @@
       require('unload')
     ])
     .init();
-
-  // var moduleNames = Object.keys(Mojular.Modules);
-  // if(moduleNames.length) {
-  //   var moduleList = moduleNames.map(function(i) {
-  //     return '• ' + i;
-  //   }).join('\n').replace(/^\s+/, '');
-  //   console.log('The following modules are loaded:\n' + moduleList);
-  // }
 }());
--- a/mtp_cashbook/templates/cashbook/transactions_history.html
+++ b/mtp_cashbook/templates/cashbook/transactions_history.html
@@ -41,15 +41,15 @@
     {% endif %}
 
     <div class="grid-row">
-    <div id="form-group-search" class="form-group ">
+    <div id="form-group-search" class="form-group {% if form.search.errors %}error{% endif %}">
         <div class="column-full">
           <label for="id_search" id="id_search-label" class="form-label">
             {% trans 'Prisoner name, prisoner number or sender name' %}
-            {% for error in field.errors %}
+            {% for error in form.search.errors %}
               <span class="error-message">{{ error }}</span>
             {% endfor %}
           </label>
-          <input class=" form-control" id="id_search" name="search" type="text" />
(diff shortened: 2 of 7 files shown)


mtp_cashbook/assets-src/javascripts/main.js
```diff
@@ -22,11 +23,11 @@
...
-    }).join('\n').replace(/^\s+/, '');
-    console.log('The following modules are loaded:\n' + moduleList);
-  }
+  // var moduleNames = Object.keys(Mojular.Modules);
+  // if(moduleNames.length) {
+  //   var moduleList = moduleNames.map(function(i) {
+  //     return '• ' + i;
+  //   }).join('\n').replace(/^\s+/, '');
+  //   console.log('The following modules are loaded:\n' + moduleList);
+  // }
```
SteveMarshall: If we're not using code, we should remove it, not comment it out. That's what git history is for :wink:

mtp_cashbook/templates/cashbook/transactions_history.html
```diff
@@ -40,18 +40,45 @@ <h1 class="error-summary-heading heading-medium" id="error-summary-heading">{% t
...
+          </label>
+          <input class=" form-control" id="id_search" name="search" type="text" />
+        </div>
       </div>
-    {% endfor %}
-
+    </div>
+
+    <div class="grid-row">
+      <div class="column-full"><label>Date received<label></div>
```
SteveMarshall: Missing `trans`.

mtp_cashbook/templates/cashbook/transactions_history.html
```diff
@@ -77,11 +104,11 @@ <h3 class="HistoryHeader {% if transaction_status == 'uncredited' %}HistoryHeade
             <a class="u-nojs-hidden print-hidden HistoryHeader-aside ShowHide js-ShowHide" href="#HistoryBatch-{{ forloop.parentloop.counter }}-{{ forloop.counter }}">
               <span class="ShowHide-hide">
                 {% trans 'Collapse' %}
-                <span class="ShowHide-icon">&minus;</span>
+                <span class="ShowHide-icon">&ndash;</span>
               </span>
               <span class="ShowHide-show">
                 {% trans 'Collapsed' %}
-                <span class="ShowHide-icon">&plus;</span>
+                <span class="ShowHide-icon">+</span>
```
SteveMarshall: Shouldn't these be images with appropriate `alt`s? (The en dash, I think, won't be read out by screen readers meaningfully, for example?)

Also… ultimately, they should be added by JS because they don't do anything without it.

mtp_cashbook/templates/cashbook/transactions_history.html
```diff
@@ -40,18 +40,45 @@ <h1 class="error-summary-heading heading-medium" id="error-summary-heading">{% t
...
+    <div class="grid-row">
+      <div class="column-full"><label>Date received<label></div>
+      <div class="column-half">
+        <div id="form-group-start" class="form-group ">
+          <label for="id_start" id="id_start-label" class="form-label">
+            {% trans 'From' %}
+            {% for error in field.errors %}
+              <span class="error-message">{{ error }}</span>
+            {% endfor %}
+          </label>
```
SteveMarshall: How will these `trans` blocks and `{{ error }}`s end up rendering? Should the `{{ errors }}` be interpolated into the translation, perhaps?

mtp_cashbook/templates/cashbook/transactions_history.html
```diff
@@ -40,18 +40,45 @@ <h1 class="error-summary-heading heading-medium" id="error-summary-heading">{% t
...
-          {% for error in field.errors %}
-            <span class="error-message">{{ error }}</span>
-          {% endfor %}
-        </label>
-        {{ field }}
+    <div class="grid-row">
+    <div id="form-group-search" class="form-group ">
+        <div class="column-full">
+          <label for="id_search" id="id_search-label" class="form-label">
+            {% trans 'Prisoner name, prisoner number or sender name' %}
```
SteveMarshall: What's the advantage of having these here instead of using the `field.label` as it was? This seems more brittle to me…

mtp_cashbook/templates/cashbook/transactions_history.html
maxf: Agreed. Given that this is going to UR tomorrow, I suggest we should leave as is for now

mtp_cashbook/templates/cashbook/transactions_history.html
maxf: The problem with how it was before is that the backend generates the form, so there's no control on style. Worse, there's non-translatable text in forms.py

mtp_cashbook/templates/cashbook/transactions_history.html
maxf: Yes, we should do like below.

mtp_cashbook/templates/cashbook/transactions_history.html
ushkarev: actually they were translatable (the `_()` wrapper), but we agreed that it's easier to keep the copy in one logical place (the template) and the django form fields really restrict the html output you get.

ushkarev:
v minor, but webpack is complaining about `console` in line 1 of `mtp_cashbook/assets-src/javascripts/main.js`

mtp_cashbook/templates/cashbook/transactions_history.html
```diff
@@ -40,18 +40,45 @@ <h1 class="error-summary-heading heading-medium" id="error-summary-heading">{% t
...
+      <div class="column-full"><label>{% trans 'Date received' %}</label></div>
+      <div class="column-half">
+        <div id="form-group-start" class="form-group ">
+          <label for="id_start" id="id_start-label" class="form-label">
+            {% trans 'From' %}
+            {% for error in field.errors %}
+              <span class="error-message">{{ error }}</span>
+            {% endfor %}
+          </label>
+          <input class=" form-control js-FormDateControl" id="id_start" name="start" type="text" value="19/11/2015" />
```
ushkarev: `value` attribute is wrong… shall i fix-up?

PR #113, opened 2015-12-02
Add a makefile and a script to do the same as gulp, but simpler

maxf:
- the makefile should be launched for various build and run tasks
- the script shouldn't normally be launched on its own but it started by the makefile
- jshint slightly corrected to allow JS files to pass linting

[This is an early version, missing a few things and not field tested much, but doesn't
interfere with the gulp process and early feedback is welcome]


run.sh
```diff
@@ -0,0 +1,20 @@
...
+  if [ ! -d node_modules ]; then
+    echo "The installation process is about to start. It usually takes a while."
+    echo "The only thing that this script doesn't do is set up the API. While"
+    echo "installation is running, head to https://github.com/ministryofjustice/money-to-prisoners-api"
+    echo "to find out how to run it."
+    echo "Press Return to continue"
+    read
+    npm install
+  fi
+  make -f node_modules/money-to-prisoners-common/Makefile $1 app=mtp_cashbook port=8001
```
ushkarev: maybe pass in the app parameter as:

    app=`basename $PWD`
    make -f node_modules/money-to-prisoners-common/Makefile $1 app=$app port=8001

because the folder name can vary, e.g. my folders are the same names as the repos in github

run.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+if [ "$#" -ne 1 ]; then
+  echo "Usage: $0 [start|watch|serve|clean|all] app=<app_name> port=8001"
```
ushkarev: the `app=<app_name> port=8001` part wouldn't pass the parameters through to make, would it?

run.sh
maxf: This is the name of the django app folder, not the repo's, so needs to be hardwired.

run.sh
maxf: true. Fixed.

PR #118, opened 2015-12-04
Improve printed pages

maxf:
- remove links and what's not needed
- add specific print-only boxes
- make layout printer-friendly

[Updates: #109440804]

### DIFF
--- a/mtp_cashbook/templates/cashbook/transaction_batch_list.html
+++ b/mtp_cashbook/templates/cashbook/transaction_batch_list.html
@@ -56,6 +56,10 @@
 
         {% include "cashbook/includes/batch_buttons.html" %}
 
+        <div class="print-box">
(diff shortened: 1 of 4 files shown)


mtp_cashbook/templates/cashbook/transaction_batch_list.html
```diff
@@ -56,6 +56,10 @@ <h1 class="error-summary-heading heading-medium" id="error-summary-heading">{% t
 
         {% include "cashbook/includes/batch_buttons.html" %}
 
+        <div class="print-box">
+          Remember to click 'Done' in the digital cashbook when you’ve processed credits in NOMIS.
```
ushkarev: we should probably wrap this in `{% trans '' %}` for consistency…

PR #119, opened 2015-12-08
Remove the left margin on the Done button line

maxf:
[Closes #109435906]

### DIFF
--- a/mtp_cashbook/assets-src/stylesheets/views/_batch.scss
+++ b/mtp_cashbook/assets-src/stylesheets/views/_batch.scss
@@ -71,7 +71,7 @@
   }
 
   .batch-buttons {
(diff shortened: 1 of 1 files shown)


mtp_cashbook/assets-src/stylesheets/views/_batch.scss
```diff
@@ -71,7 +71,7 @@
   }
 
   .batch-buttons {
-    margin: 1em 1em;
+    margin: 1em 0 1em 0;
```
SteveMarshall: Any reason not to use the shorthands, like `1em 0`?

maxf: No, no reason. I'll shorten when amending the file next time.

PR #128, opened 2015-12-15
Improved credit printout

maxf:


mtp_cashbook/assets-src/javascripts/modules/running-total.js
```diff
@@ -0,0 +1,67 @@
+// Running Total module
+/* globals require, exports */
+'use strict';
```
ushkarev: lint says `Use the function form of "use strict".` – but the node was is to have it at top level, right?

maxf: Correct. Non-function form should be allowed.
The setting for jshint to stop complaining doesn't seem to be working.

maxf:
Yep. Good point. Changed

PR #132, opened 2015-12-18
Don't pin the version number

maxf:
We can then avoid trickling down dependencies in all 3 applications.
This is considered safe as we control the common module


ushkarev:
and the idea is to re-pin later on in the development timeline once the dependencies are stabilised

PR #149, opened 2016-01-11
Update run script to share more functionality and allow parameter passing

ushkarev:
depends on [mtp-common#47](https://github.com/ministryofjustice/money-to-prisoners-common/pull/47)

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -7,25 +7,13 @@ if [ ! -d node_modules ]; then
   echo "The only thing that this script doesn't do is set up the API. While"
   echo "installation is running, head to https://github.com/ministryofjustice/money-to-prisoners-api"
   echo "to find out how to run it."
-
-  read -p "Proceed with installation? [Y/n]: " PROCEED
-  case ${PROCEED} in
-    [nN]*) exit 1;;
-  esac
-
   npm install
 fi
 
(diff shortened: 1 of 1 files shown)


run.sh
```diff
@@ -1,20 +1,31 @@
...
+  echo "The installation process is about to start. It usually takes a while."
+  echo "The only thing that this script doesn't do is set up the API. While"
+  echo "installation is running, head to https://github.com/ministryofjustice/money-to-prisoners-api"
+  echo "to find out how to run it."
+
+  read -p "Proceed with installation? [Y/n]: " PROCEED
+  case ${PROCEED} in
+    [nN]*) exit 1;;
+  esac
+
```
maxf: (not related to this PR) This should be able to run non-interactively in case this script is run in a docker container

maxf:
This is shifting the main tasks away to the makefile and into the tasks. Normally make is what you run , and it calls relevant scripts, tasks.sh exists to make the makefile visibly more about the dependencies and recipes (although it has been argued that the commands should be put inline directly and not in a tasks.sh file). The reason why run.sh exists is that there's already a makefile which I didn't want to touch until the new makefile could be integrated into the existing one, which would have removed the need for run.sh.

PR #150, opened 2016-01-12
Reflect the current processes

maxf:
This is currently mostly about front-end commands. Feel free to let me know if anything's missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,115 +5,72 @@
 
 The Cashbook UI for the Money to Prisoners Project
 
-## Dependencies
-### Docker
-To run this project locally you need to have
-[Docker](http://docs.docker.com/installation/mac/) and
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,115 +5,72 @@
...
+- *test-prison-1 / test-prison-1* for Prison 1
+- *test-prison-2 / test-prison-2* for Prison 2
 
-### Run a development Server
+## Developing
 
-In a terminal `cd` into the directory you checked this project out into, then
+With the `run.sh` command, you can run a browser-sync server, and get the assets
+to automatically recompile when changes are made, run `./run.sh serve` instead of
+`./run.sh start`. The server is then available at [http://localhost:3000](http://localhost:3000)
```
ushkarev: actually, the port will vary per-application so maybe you can say that it will print the url that it's serving on

PR #151, opened 2016-01-12
Tell GA the number of events submitted

maxf:
And fix a markup bug which messed up the credit form validation


mtp_cashbook/assets-src/javascripts/modules/batch-validation.js
```diff
@@ -60,7 +60,10 @@ exports.BatchValidation = {
     var checkedValid = this._allChecked();
     var numChecked = this._numChecked();
 
+    ga && ga('send', 'event', 'credits', 'submit', numChecked);
```
ushkarev: do you think the total quantity of credits should be somehow included? in case there weren't enough originally to make a batch of 20 (which i think is the default)

PR #185, opened 2016-03-18
Place date fields on the side for shorter form

maxf:


ushkarev:
may just be me not compiling the assets correctly, but there's some strange alignment going on…

PR #210, opened 2016-04-14
Replace error summary and fields with shared templates from MTP-utils

ushkarev:
will not pass tests until [mtp-utils#13](https://github.com/ministryofjustice/money-to-prisoners-utils/pull/13) is merged and on pypi

### DIFF
--- a/mtp_cashbook/apps/cashbook/forms.py
+++ b/mtp_cashbook/apps/cashbook/forms.py
@@ -40,7 +40,7 @@ class ProcessTransactionBatchForm(GARequestErrorReportingMixin, forms.Form):
         transactions = self.cleaned_data.get('transactions', [])
         discard = self.data.get('discard') == '1'
         if not transactions and not discard:
-            raise forms.ValidationError(ugettext('Only click ‘Done’ when you’ve selected credits'))
+            self.add_error(None, ugettext('Only click ‘Done’ when you’ve selected credits'))
         if discard:
             transactions = []
         return transactions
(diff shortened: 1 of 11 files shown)


mtp_cashbook/apps/cashbook/forms.py
```diff
@@ -40,7 +40,7 @@ def clean_transactions(self):
         transactions = self.cleaned_data.get('transactions', [])
         discard = self.data.get('discard') == '1'
         if not transactions and not discard:
-            raise forms.ValidationError(ugettext('Only click ‘Done’ when you’ve selected credits'))
+            self.add_error(None, ugettext('Only click ‘Done’ when you’ve selected credits'))
```
ushkarev: set a non-field-specific error message because a linkable form field does not really exist on the page

mtp_cashbook/apps/cashbook/forms.py
```diff
@@ -10,9 +10,12 @@
 from form_error_reporting import GARequestErrorReportingMixin
 from moj_auth.api_client import get_connection
 from mtp_utils.api import retrieve_all_pages
+from mtp_utils.forms import replace_default_error_messages
 
 from .form_fields import MtpTextInput, MtpDateInput
 
+replace_default_error_messages()
```
ibrechin: If we are really replacing default messages everywhere, it might be better to put this in ```settings``` or in the ```__init__.py``` of the base package rather than in every ```forms.py```. I mean I'm not sure if any of our client apps actually have more than one ```forms.py``` but you know.

ushkarev:
retest this please

PR #231, opened 2016-04-29
Introduce new script to run dev tasks

maxf:


run.sh
```diff
@@ -0,0 +1,83 @@
...
+        ;;
+    esac
+    shift
+done
+
+COMMAND=$1
+
+case "$COMMAND" in
+    clean)
+        git clean -dfX
```
ibrechin: I don't think 'clean' should do a git clean, that seems like overkill. Should only target the generated files.

run.sh
```diff
@@ -0,0 +1,83 @@
...
+        PORT=8001
+        BROWSERSYNC_PORT=3001
+        BROWSERSYNC_UI_PORT=3031
+        VIRTUAL_ENV_DIR=venv
+        NODE_MODULES=node_modules
+        MTP_COMMON_NODE_MODULES=$NODE_MODULES/money-to-prisoners-common
+
+        echo "Creating Python virtual environment"
+        virtualenv --python=python3 $VIRTUAL_ENV_DIR >/dev/null
+        source $VIRTUAL_ENV_DIR/bin/activate >/dev/null
```
ibrechin: Can we put the preceding two lines inside a ```if ! [ -z "$VIRTUAL_ENV" ]```?

run.sh
```diff
@@ -0,0 +1,83 @@
...
+        APP=cashbook
+        PORT=8001
+        BROWSERSYNC_PORT=3001
+        BROWSERSYNC_UI_PORT=3031
+        VIRTUAL_ENV_DIR=venv
+        NODE_MODULES=node_modules
+        MTP_COMMON_NODE_MODULES=$NODE_MODULES/money-to-prisoners-common
+
+        echo "Creating Python virtual environment"
+        virtualenv --python=python3 $VIRTUAL_ENV_DIR >/dev/null
```
ibrechin: Maybe a ```[ -d $VIRTUAL_ENV_DIR ]``` here?

run.sh
maxf: The current Makefile already does a git clean. It only removes the generated files, too.

run.sh
maxf: ok

run.sh
maxf: Should be VIRTUAL_ENV_DIR anyway and that's covered with your -d above

ibrechin: No, I mean $VIRTUAL_ENV. As in, if there is a virtual env already active, don't try and generate the venv/ dir.

run.sh
ibrechin: In that case the current Makefile needs to be changed as well. As is, this command will delete my settings/local.py and my project files.

PR #251, opened 2016-06-13
Tag Zendesk feedback tickets with prison(s)

ushkarev:


mtp_cashbook/apps/feedback/forms.py
```diff
@@ -0,0 +1,17 @@
...
+from slumber.exceptions import HttpClientError
+from zendesk_tickets.forms import EmailTicketForm
+
+
+class PrisonTicketForm(EmailTicketForm):
+    def submit_ticket(self, request, subject, tags, ticket_template_name, **kwargs):
+        try:
+            client = get_connection(request)
+            prisons = sorted(map(lambda prison: prison['name'], client.prisons.get()['results']))
+            subject += ' (%s)' % ', '.join(prisons)
```
SteveMarshall: This might be clearer if the `prisons` variable had a clearer name. Maybe `affected_prisons`?

I might also be tempted to make the subject assignment something like this, so the ultimate structure of the subject is a bit more explicit:

```python
subject = '{subject} ({affected_prisons})' % {
    'subject': subject,
    'affected_prisons': affected_prisons
}
```

ushkarev: but `{subject} ({affected_prisons})` is just the same as appending…
ps: the list of prisons is the ones the user has access to

SteveMarshall: I'd argue that the format string makes the structure more explicit; `+= ' (%s)'`  is ever so slightly harder to parse, and will be problematic if we ever end up translating the interface into Welsh. (Is that a possibility? I have no idea…)

`accessible_prisons` seems like an appropriate name to me, though.

ushkarev: we wouldn't need to translate this to welsh; it's a ticket that gets created in zendesk – would be odd to arbitrarily make some tickets in welsh. i still do not agree that the above is any clearer, but since it makes you happy…

SteveMarshall: 😍

PR #313, opened 2016-10-31
New credits tweaks

vickymg:


mtp_cashbook/templates/cashbook/credit_batch_list.html
```diff
@@ -80,7 +80,7 @@ <h2 class="heading-small">
             </div>
 
             <div class="column-one-third">
-              <p>{% trans 'Entered into NOMIS by you' %}(<span class="count-checked-checkboxes">0</span>)</p>
+              <p>{% trans 'Entered into NOMIS by you ' %}(<span class="count-checked-checkboxes">0</span>)</p>
```
ushkarev: i think the space ought to be outside the translation string so that translators don't have to always make sure to end with one. i.e. `you' %} (<sp`

PR #324, opened 2016-11-08
Add align-totals script to line up totals with amounts column

vickymg:


mtp_cashbook/assets-src/javascripts/modules/align-totals.js
```diff
@@ -0,0 +1,30 @@
...
+  init: function () {
+    this.cacheEls();
+    this.bindEvents();
+  },
+
+  cacheEls: function () {
+    this.$window = $(window);
+    this.amount = $('td.numeric:first');
+    this.ctrlTotal = $('.ctrl-total');
+    this.entTotal = $('.entered-total');
```
ushkarev: might be worth having some sort of check that one of these exists so that the event listeners don't get attached to window on all pages.

PR #354, opened 2017-02-27
Update django version to 1.10

ibrechin:


mtp_cashbook/apps/cashbook/tests/test_functional.py
```diff
@@ -316,12 +316,12 @@ def test_journey_11(self):
 
     def test_journey_12(self):
         self.login('test-prison-1', 'test-prison-1')
-        self.click_on_text('View')
+        self.click_on_text('Return these to new credits')
         self.click_on_text('Release')
 
     def test_journey_13(self):
         self.login('test-prison-1', 'test-prison-1')
-        self.click_on_text('View')
```
ushkarev: i totally changed that in another PR… but only for phantomjs-driven tests :)

PR #359, opened 2017-03-24
Send email notification to sender when prisoner is credited

ibrechin:
This only applies to those credited via NOMIS integration, as
otherwise we cannot be certain that the prisoners actually received
the money.


ibrechin:
retest this please

PR #362, opened 2017-03-30
Fix JS bug with running total summation

ibrechin:
Larger credit amounts with a comma in would be parsed as if the
comma were a decimal i.e. £1,000 would be added to the running
total as £1.


ibrechin:
retest this please

PR #385, opened 2017-06-27
Docker upgrades

ushkarev:
This updates
• the base container to Ubuntu Xenial
• node.js to version 8
• python to version 3.5

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,36 +1,38 @@
-FROM ubuntu:trusty
+FROM buildpack-deps:xenial
 
 # setup environment
-RUN echo "Europe/London" > /etc/timezone && dpkg-reconfigure --frontend noninteractive tzdata
-RUN locale-gen "en_GB.UTF-8"
-ENV LC_CTYPE=en_GB.UTF-8
-ENV TERM=xterm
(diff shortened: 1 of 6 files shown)


mtp_cashbook/assets-src/javascripts/modules/confirm-manual.js
```diff
@@ -31,13 +31,13 @@ exports.ConfirmManual = {
       return;
     }
 
-    var credit_id = $el.data('credit-id');
+    var creditID = $el.data('credit-id');
```
ibrechin: Personally I prefer ```creditId``` if doing camel case.

Dockerfile
```diff
@@ -1,37 +1,39 @@
...
-RUN mkdir -p mtp_cashbook/assets
-RUN mkdir -p mtp_cashbook/assets-static
-RUN mkdir -p static
-RUN mkdir -p media
-RUN mkdir -p spooler
-RUN chown www-data:www-data spooler
-
-# install python
-RUN apt-get install -y python3-all python3-all-dev python3-setuptools python3-pip python-pip python3.4-venv
+RUN set -ex; mkdir -p \
```
ibrechin: Personally I think trailing backslashes for newlines look super gross.

PR #410, opened 2017-11-27
Disbursements

ibrechin:


mtp_cashbook/apps/disbursements/forms.py
```diff
@@ -0,0 +1,183 @@
...
+            prisoner_number = prisoner_number.upper()
+        return prisoner_number
+
+    def clean(self):
+        prisoner_number = self.cleaned_data.get('prisoner_number')
+        session = get_api_session(self.request)
+        try:
+            prisoner = session.get(
+                '/prisoner_locations/{prisoner_number}/'.format(
+                    prisoner_number=prisoner_number
```
ushkarev: we should url-encode `prisoner_number` to prevent api endpoint traversal

mtp_cashbook/apps/disbursements/forms.py
```diff
@@ -0,0 +1,183 @@
...
+            'Use NOMIS to move money from other accounts into the private '
+            'account, then click ’Update balances’'),
+    }
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        from .views import PrisonerView
+        try:
+            balances = nomis.get_account_balances(
+                self.previous_form_data[PrisonerView.url_name]['prison'],
```
ushkarev: `get_account_balances` does not url-encode so either we should change common or we escape here

PR #412, opened 2017-11-28
Add navigation between sub-apps

ushkarev:


mtp_cashbook/apps/cashbook/context_processors.py
```diff
@@ -1,14 +1,23 @@
...
 
 def govuk_localisation(request):
+    global_header_text = gettext('Manage prisoners’ money')
+    if request.disbursements_available and hasattr(request, 'proposition_app'):
+        # NB: ensure proposition_app variables are html-safe
+        global_header_text += '</a> ' \
+                              '<span class="mtp-header-app">' \
+                              '<a href="%(url)s">%(name)s</a>' \
+                              '</span>' % request.proposition_app
+        global_header_text = mark_safe(global_header_text)
```
ibrechin: gross

ushkarev: the only way i could think to add an extra element without resorting to JS

PR #418, opened 2017-12-06
Improve sort code display

ushkarev:


mtp_cashbook/apps/disbursements/forms.py
```diff
@@ -189,13 +190,25 @@ class RecipientContactForm(DisbursementForm):
...
 
+    def clean_postcode(self):
+        postcode = self.cleaned_data.get('postcode')  # type: str
+        if postcode:
+            postcode = postcode.upper()
+        return postcode
+
+
+validate_account_number = RegexValidator(r'^\d{8}$', message=_('Invalid account number'))
+validate_sort_code = RegexValidator(r'^\d\d-?\d\d-?\d\d$', message=_('Invalid sort code'))
```
ibrechin: I realise this is just placeholder copy but might it be worth putting something slightly more explanatory in, just in case we forget to change it?

PR #427, opened 2018-01-11
Confirm disbursements

ibrechin:


mtp_cashbook/apps/disbursements/utils.py
```diff
@@ -0,0 +1,43 @@
...
+        location = nomis.get_location(disbursement['prisoner_number'])
+
+        viability['prisoner_moved'] = (
+            disbursement['prison'] != location['nomis_id']
+        )
+    except RequestException:
+        pass
+
+    viability['self_own'] = (
+        disbursement['log_set'][-1]['user']['username'] == request.user.username
```
ushkarev: regarding `log_set` ordering… the default is by `pk`, but i wonder if `created` would be better…?

also the various viabilities could be in the form `any(… for log_set in disbursement['log_set'] if log_set[…] == …)` – not asking for a change, just wondering if it'd be clearer or more generic

PR #442, opened 2018-03-13
Error when building society account details entered for disbursement

ibrechin:
Disbursements to building societies are not currently supported.


mtp_cashbook/apps/disbursements/forms.py
```diff
@@ -242,6 +243,17 @@ class RecipientBankAccountForm(DisbursementForm):
         validators=[validate_sort_code],
     )
 
+    def clean(self):
+        cleaned_data = super().clean()
+        if roll_number_required(
+            cleaned_data.get('sort_code'), cleaned_data.get('account_number')
+        ):
+            raise forms.ValidationError(_(
```
ushkarev: could we have these as `gettext()+gettext()` to separate the sentences?

PR #446, opened 2018-04-13
Postcode lookup

ibrechin:
Add address lookup facility to disbursement creation flow.


ibrechin:
I can give you a token if you want to test it out but don't merge anything yet as we still don't have a proper account.

mtp_cashbook/assets-src/javascripts/modules/address-picker.js
```diff
@@ -0,0 +1,30 @@
...
+  bindAddressSelection: function () {
+    var addressSelect = $('#address-select');
+
+    var addressLine1 = $('#id_address_line1');
+    var addressLine2 = $('#id_address_line2');
+    var city = $('#id_city');
+    var postcode = $('#id_postcode');
+
+    function update () {
+      if (addressSelect.val() > 0) {
```
ushkarev: the value is a string, so i'd personally use `if (addressSelect.val())` and remove `value="0"`

ushkarev:
functional tests would need updating too

mtp_cashbook/apps/disbursements/utils.py
```diff
@@ -49,3 +50,77 @@ def get_disbursement_viability(request, disbursement):
...
+                street_number = building_number or street
+
+            locality = build_address_line(
+                address_data, 'DOUBLE_DEPENDENT_LOCALITY', 'DEPENDENT_LOCALITY'
+            )
+
+            line_1 = ''
+            line_2 = ''
+
+            for item in [business, po_box, building, street_number, locality]:
```
ushkarev: maybe `line1, *line2 = list(filter(None, [business, po_box, building, street_number, locality])); line2 = ', '.join(line2)` or is that going too far? it's just that we're like 6 indents deep…

PR #451, opened 2018-05-04
Allow disbursements to companies

Katy600:


mtp_cashbook/apps/disbursements/views.py
```diff
@@ -437,8 +445,8 @@ def create_nomis_transaction(self, form):
                 record_id='d%s' % disbursement['id'],
                 description='Sent to {recipient_first_name} {recipient_last_name}'.format(
                     recipient_first_name=disbursement['recipient_first_name'],
-                    recipient_last_name=disbursement['recipient_last_name']
-                ),
+                    recipient_last_name=disbursement['recipient_last_name'],
+                ).replace('  ', ' '),
```
ibrechin: Is this replacing double spaces with single spaces? I would think it would be better to trim the inputs. Or just don't worry about it, why are people putting spaces at the start/end of names anyway?

mtp_cashbook/apps/disbursements/views.py
```diff
@@ -576,19 +593,30 @@ def get_context_data(self, **kwargs):
...
+        elif disbursement_data.get('recipient_type') == 'company':
+            disbursement_data['recipient_is_company'] = True
+            disbursement_data['recipient_last_name'] = disbursement_data['recipient_company_name']
+            disbursement_data['recipient_first_name'] = ''
+        if 'recipient_type' in disbursement_data:
+            del disbursement_data['recipient_type']
+        if 'recipient_company_name' in disbursement_data:
+            del disbursement_data['recipient_company_name']
+        if 'remittance' in disbursement_data:
+            del disbursement_data['remittance']
```
ibrechin: I would think all this stuff should go in the specific `get_upload_payload` of the form or view, rather than in the base class.

mtp_cashbook/apps/disbursements/views.py
```diff
@@ -555,13 +563,22 @@ def dispatch(self, request, **kwargs):
...
+            disbursement_data = self.disbursement.copy()
+            if disbursement_data.get('recipient_is_company'):
+                disbursement_data['recipient_type'] = 'company'
+                disbursement_data['recipient_company_name'] = disbursement_data['recipient_last_name']
+                disbursement_data['recipient_first_name'] = ''
+                disbursement_data['recipient_last_name'] = ''
+            else:
+                disbursement_data['recipient_type'] = 'person'
+                disbursement_data['recipient_company_name'] = ''
+            disbursement_data['remittance'] = 'yes' if disbursement_data.get('remittance_description') else 'no'
```
ibrechin: This should probably also be in the form class or view subclass. It's in a loop here, so it's going to be doing it over and over again which seems weird.

PR #475, opened 2018-10-16
Promotion of the security tool on the cashbook login page

Katy600:


mtp_cashbook/templates/mtp_auth/login.html
```diff
@@ -61,6 +61,11 @@ <h2 class="heading-medium">{% trans 'Need access?' %}</h2>
             {% trans 'Request access' %}
           </a>
         </p>
+        <p>
+          {% blocktrans trimmed %}
+            Intelligence staff need to log into the <a href=" https://noms-ops.moneytoprisoners.dsd.io/">prisoner money intelligence tool</a> instead.
```
ushkarev: this is a hard-coded url which is a bad idea cos we've got different versions of this site. see the footer link about how it ought to be done. also, that space at the beginning of the link is an error, pls watch out for random spaces

PR #525, opened 2020-08-03
Add monitoring to flag when a prisoner's acount balance exceeds cap

ushkarev:
…by checking NOMIS when a credit is credited and comparing to threshold.

[MTP-1479](https://dsdmoj.atlassian.net/browse/MTP-1479)

The idea being to get a qualitative measure of the frequency with which credits being credited actually breach the cap. We expect these to be very rare so this is simply flagged as an error each time it occurs rather than measured quantitatively.


mtp_cashbook/apps/cashbook/tasks.py
```diff
@@ -87,3 +95,26 @@ def credit_individual_credit_to_nomis(user, user_session, credit_id, credit):
...
+
+
+NOMIS_ACCOUNTS = {'cash', 'spends', 'savings'}
+
+
+@spoolable()
+def check_balance_is_below_cap(prison, prisoner_number):
+    # NB: balances are not known in private estate currently, but cashbook is not used in there
+    try:
+        nomis_account_balances = nomis.get_account_balances(prison, prisoner_number)
```
Staberinde, line 107: This is pretty similar to the checking done in the API, I wonder if we should make this a common method between the two?

mtp_cashbook/apps/cashbook/tests/test_views.py
```diff
@@ -232,6 +233,127 @@ def test_new_credits_submit(self, mock_create_transaction):
                 '£52.00' in mail.outbox[1].body and '£45.00' in mail.outbox[0].body
             )
 
+    @override_settings(
+        ENVIRONMENT='prod',  # because non-prod environments don't send to .local
+        PRISONER_CAPPING_ENABLED=True,
+        PRISONER_CAPPING_THRESHOLD_IN_POUNDS=100,
+    )
+    @mock.patch('cashbook.tasks.logger')
+    @mock.patch('cashbook.tasks.nomis')
```
Staberinde, line 242: Seems kinda weird having `RequestsMock` mock everything except the nomis requests. Any reason for this?

mtp_cashbook/apps/cashbook/tests/test_views.py
```diff
@@ -232,6 +233,127 @@ def test_new_credits_submit(self, mock_create_transaction):
...
+            self.login()
+            response = self.client.post(
+                self.url,
+                data={'credits': [1, 2], 'submit_new': None},
+                follow=True
+            )
+
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, '2 credits sent to NOMIS')
+        logger_error_messages = [call[0][0] for call in mock_logger.error.call_args_list]
```
Staberinde, line 353: I really don't like asserting against calls to the logging framework. Logging is ancillary functionality, there should not be a dependence on the veracity of the code flow based on what the logger sees. I would need to look at the code a bit closer but my hunch is that this functionality can be replicated by asserting against a spy function,

ushkarev: the purpose of this feature is to log an error, so really that _is_ the thing we want to assert is happening

mtp_cashbook/apps/cashbook/tasks.py
ushkarev: it's a combination of both the api and the send-money code so it can't entirely be replaced with a call to the api. it is a slight repetition, i agree. when we decide what to do with the outstanding issues regarding capping, we could maybe consolidate all of that within mtp-common?

mtp_cashbook/apps/cashbook/tests/test_views.py
ushkarev: it's the same as the other tests in this file. also, this test doesn't return a pre-set value for NOMIS requests, it actually handles the data passed in.

mtp_cashbook/apps/cashbook/tasks.py
Staberinde: Sure, I don't think we need to consolidate it as part of this ticket

mtp_cashbook/apps/cashbook/tests/test_views.py
Staberinde: Yeah it's true that it's similar to the rest of the file, so I won't block on it

mtp_cashbook/apps/cashbook/tests/test_views.py
Staberinde: I guess that's true in this case, since there's a functional requirement for the logging message to be added (as opposed to non-functional) we need to test that requirement.

PR #530, opened 2020-10-01
MTP-1589 Add banners warning of policy change

mattbee:
Jira:  https://dsdmoj.atlassian.net/browse/MTP-1589

To inform staff members using the cashbook of upcoming policy changes
that affect credits and disbursements, we are including an overview
banner on the dashboard and more details on the credit/disbursement
page so staff know what's coming.

I templated the banners as one is repeated on 2 credit pages and as they
will change on 2nd November to announce the policy change, it makes the
changes needed easy to find and easy to remove when the time comes.


mtp_cashbook/templates/policy-change-warning-credits.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n %}
+
+<div class="mtp-notification mtp-notification--info">
+  <h2 class="mtp-notification__headline">
+    {% trans 'No more bank transfers' %}
+  </h2>
+  <div class="mtp-notification__message" style="display:block">
```
mattbee, line 7: I don't like this inline style, however it's been done before and would need a tmp-common change to remove it in a sensible way, so living with it for now, is that fair?

ushkarev: yeah it's been done for "temporary" banners in send-money. so far, i don't think we've got any that are expected to be around after nov 2, right?

ushkarev: actually, i can't remember if some variation of this banner is supposed to remain after nov 2. if so, you could add a new sibling class to `mtp-notification__message` to force the banner open if you wanted? (either in common or here)

ushkarev (dismissed):
don't have any real issues with this if it's relatively temporary. i think if some form of banner will exist forever, might be good to have design input cos it takes up a lot of space and might feel annoying. so many options tho… e.g. have it collapsed by default (already possible) or have it be dismiss-able with a close button (not currently possible) or have a smaller banner that links out to new pages…

mtp_cashbook/templates/policy-change-warning-credits.html
mattbee: Actually, looking at the CSS I can just include the class that is toggled by JS and force the box to be open without the inline style, updated.

ushkarev: oh cool does that work? i thought i tried it in send-money, but it looked funny; clearly didn't try hard enough!

ushkarev (dismissed):

ushkarev (approved):

PR #534, opened 2020-10-28
Upgrade Django to 2.2 & fix tests

mattbee:
Main change is passing an empty string in tests instead of None. An alternative option would be changing the content type on the form to `application/json` but as a HTML form would not send null values, I think an empty string will serve the same purpose as None for the tests. Is that fair?


ushkarev:
those tests that broke actually look weird to me. i think `None` was the wrong this to do in the first place.

just taking the first as an example, the form being submitted would actually send `'submit'` or `'override'`:
https://github.com/ministryofjustice/money-to-prisoners-cashbook/blob/master/mtp_cashbook/templates/cashbook/new_credits.html#L182
https://github.com/ministryofjustice/money-to-prisoners-cashbook/blob/master/mtp_cashbook/templates/cashbook/new_credits.html#L353

mattbee:
> those tests that broke actually look weird to me. i think `None` was the wrong thing to do in the first place cos it's not what the form submitted.
> 
> just taking the first as an example, the form being submitted would actually send `'submit'` or `'override'`:
> https://github.com/ministryofjustice/money-to-prisoners-cashbook/blob/master/mtp_cashbook/templates/cashbook/new_credits.html#L182
> https://github.com/ministryofjustice/money-to-prisoners-cashbook/blob/master/mtp_cashbook/templates/cashbook/new_credits.html#L353
> 
> it might be fine to use empty strings, but it's more realistic to find what values are really being posted

Good point, I admit I assumed that the form names would be accurate in the tests. I'll check what is submitted and at least match the names and I think empty string would be the right thing to submit as long as those fields are correct (although if we can submit nothing at all that might be even better...)

mattbee:
> > those tests that broke actually look weird to me. i think `None` was the wrong thing to do in the first place cos it's not what the form submitted.
> > just taking the first as an example, the form being submitted would actually send `'submit'` or `'override'`:
> > https://github.com/ministryofjustice/money-to-prisoners-cashbook/blob/master/mtp_cashbook/templates/cashbook/new_credits.html#L182
> > https://github.com/ministryofjustice/money-to-prisoners-cashbook/blob/master/mtp_cashbook/templates/cashbook/new_credits.html#L353
> > it might be fine to use empty strings, but it's more realistic to find what values are really being posted
> 
> Good point, I admit I assumed that the form names would be accurate in the tests. I'll check what is submitted and at least match the names and I think empty string would be the right thing to submit as long as those fields are correct (although if we can submit nothing at all that might be even better...)

Actually, I think the form and values in the test might be needed. It looks like there are multiple forms on the page, the `submit_new` value is here: https://github.com/ministryofjustice/money-to-prisoners-cashbook/blob/c246cc4b80236512396e5a547456eae7ca30473e/mtp_cashbook/templates/cashbook/new_credits.html#L180 and that form data key existing determines how to process the form data.

ushkarev (approved):

PR #536, opened 2021-01-25
Use correct payment ref field from mtp-api's credit response

ushkarev:
`short_ref_number` was being used in place of `short_payment_ref` by mistake.

[MTP-1793](https://dsdmoj.atlassian.net/browse/MTP-1793)

### DIFF
--- a/mtp_cashbook/templates/cashbook/email/credited-confirmation.txt
+++ b/mtp_cashbook/templates/cashbook/email/credited-confirmation.txt
@@ -4,9 +4,9 @@ GOV.UK - {% trans 'Send money to someone in prison' %}
 
 {% trans 'Dear sender,' %}
 
-{% blocktrans trimmed %}
+{% if ref_number %}{% blocktrans trimmed %}
 The payment you made - confirmation number {{ ref_number }} - has now been credited to the prisoner’s account.
-{% endblocktrans %}
+{% endblocktrans %}{% else %}{% trans 'The payment you made has now been credited to the prisoner’s account.' %}{% endif %}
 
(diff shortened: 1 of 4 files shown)


mtp_cashbook/templates/cashbook/email/credited-confirmation.txt
```diff
@@ -4,9 +4,9 @@ GOV.UK - {% trans 'Send money to someone in prison' %}
 
 {% trans 'Dear sender,' %}
 
-{% blocktrans trimmed %}
+{% if ref_number %}{% blocktrans trimmed %}
 The payment you made - confirmation number {{ ref_number }} - has now been credited to the prisoner’s account.
-{% endblocktrans %}
+{% endblocktrans %}{% else %}{% trans 'The payment you made has now been credited to the prisoner’s account.' %}{% endif %}
```
Staberinde, line 9: Very minor stylistic comment, is it worth breaking this whole block out in to a multiline statement?

Staberinde (commented):
Just one comment

mtp_cashbook/templates/cashbook/email/credited-confirmation.txt
ushkarev: i agree it looks gross! but otherwise you end up with a text-only email with 3+ blank lines. probably not a big deal since i'd expect very few people to see it, i was just aiming for tidier text email rather than tidier template

ushkarev: alas you cannot do something like `{%- if -%}` in django templates the way you can in jinja2

Staberinde (approved):

PR #540, opened 2021-03-10
FAQ when user signed out

xoen:
Added an FAQ page which user can refer to when he's not signed it.
The 'Get help' link when the user is not signed it will now take them
to this page instead of directly to the "Contact us".

⚠️ **NOTE**: This is based off [the GDS Design System branch/PR](https://github.com/ministryofjustice/money-to-prisoners-cashbook/pull/537).

Ticket: https://dsdmoj.atlassian.net/browse/MTP-1834

### DIFF
--- a/mtp_cashbook/misc_views.py
+++ b/mtp_cashbook/misc_views.py
@@ -120,3 +120,17 @@ class PolicyChangeInfo(BaseView, TemplateView):
             return ['policy-change-warning.html']
         else:
             return ['policy-change-info.html']
+
+
+class FAQView(TemplateView):
+    title = _('What do you need help with?')
+    template_name = 'faq.html'
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        context['contact_us_url'] = reverse_lazy('submit_ticket')
+        context['reset_password_url'] = reverse_lazy('reset_password')
+        context['sign_up_url'] = reverse_lazy('sign-up')
+
+        return context
--- a/mtp_cashbook/misc_views.py
+++ b/mtp_cashbook/misc_views.py
@@ -129,7 +129,7 @@ class FAQView(TemplateView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        context['contact_us_url'] = reverse_lazy('submit_ticket')
+        context['breadcrumbs_back'] = reverse_lazy('home')
         context['reset_password_url'] = reverse_lazy('reset_password')
         context['sign_up_url'] = reverse_lazy('sign-up')
 
--- a/mtp_cashbook/templates/faq.html
+++ b/mtp_cashbook/templates/faq.html
@@ -1,0 +1,81 @@
+{% extends 'base.html' %}
+{% load i18n %}
+
+{% block page_title %}{{ view.title }} – {{ block.super }}{% endblock %}
+
+{% block body_classes %}{{ block.super }} mtp-credits{% endblock %}
+
+{% block content %}
+  {{ block.super }}
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      <a class="govuk-back-link" href="{% url 'home' %}">
+        {% trans 'Back' %}
+      </a>
+
+      <header>
+        <h1 class="govuk-heading-xl">{{ view.title }}</h1>
+      </header>
+
+
+      <h2 class="govuk-heading-m" id="faq-account-requested">
+        {% trans "I requested an account, but heard nothing" %}
+      </h2>
+
+      <p class="govuk-body">
+        {% blocktrans trimmed %}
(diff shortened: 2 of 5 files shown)


mtp_cashbook/misc_views.py
```diff
@@ -121,3 +121,17 @@ def get_template_names(self):
...
+
+
+class FAQView(TemplateView):
+    title = _('What do you need help with?')
+    template_name = 'faq.html'
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        context['contact_us_url'] = reverse_lazy('submit_ticket')
```
ushkarev, line 133: as this set of urls is not dynamically changed by the view, i think it's easier just to put them into the template and have this as a really dumb view. probs as `{% url ??? as link %}` above each `blocktrans`. that way, if you're looking at the template, you can see where links point to.

mtp_cashbook/templates/faq.html
```diff
@@ -0,0 +1,81 @@
...
+      </p>
+
+
+      <h2 class="govuk-heading-m" id="faq-locked-out">
+        {% trans "I’m locked out" %}
+      </h2>
+
+      <p class="govuk-body">
+        {% blocktrans trimmed %}
+          If you mis-type your password too many times, you'll be locked out. Try again in about half an hour.
```
ushkarev, line 40: in general, it's best to split separate sentences into different translation blocks. i know that we'll probably never actually turn on other languages; this would only _theoretically_ make it easier to reuse translated snippets

mtp_cashbook/templates/faq.html
```diff
@@ -0,0 +1,81 @@
...
+      </p>
+
+
+      <h2 class="govuk-heading-m" id="faq-reset-password">
+        {% trans "Reset password isn’t working" %}
+      </h2>
+
+      <p class="govuk-body">
+        {% blocktrans trimmed %}
+          To reset your password, <a href="{{ contact_us_url }}?message=My new password is [please fill in]" class="govuk-link">contact us</a>.
```
ushkarev, line 73: could you url-encode the message part of the link

mtp_cashbook/templates/faq.html
```diff
@@ -0,0 +1,81 @@
...
+      <header>
+        <h1 class="govuk-heading-xl">{{ view.title }}</h1>
+      </header>
+
+
+      <h2 class="govuk-heading-m" id="faq-account-requested">
+        {% trans "I requested an account, but heard nothing" %}
+      </h2>
+
+      <p class="govuk-body">
```
ushkarev, line 27: you don't need `govuk-body` on paragraph elements

mtp_cashbook/templates/faq.html
```diff
@@ -0,0 +1,81 @@
...
+        {% blocktrans trimmed %}
+          Either remind your business hub manager to accept your request, or <a href="{{ sign_up_url }}" class="govuk-link">request an account</a> again.
```
ushkarev, line 29: you don't need `govuk-link` on links almost anywhere (only when it's side-by-side with a button or not inside a paragraph/heading)

mtp_cashbook/templates/faq.html
xoen: Will do. I thought in this case it was a block which made sense as a whole and that's why it made sense to have in a single block. I can change 👍 

(As a amateur translator it was sometimes tricky to understand the context of certain single sentences)

ushkarev: that is indeed true. at the moment, there's no context to _any_ of our translatable strings. it's just one long meaningless table in transifex

mtp_cashbook/misc_views.py
xoen: Ah OK, that was my first implementation, I only moved because it was used twice and I thought it would be nice to avoid the repetition. But I can move back, no problem at all ☺️

mtp_cashbook/templates/faq.html
```diff
@@ -0,0 +1,92 @@
...
+{% block body_classes %}{{ block.super }} mtp-credits{% endblock %}
+
+{% block content %}
+  {{ block.super }}
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      <a class="govuk-back-link" href="{% url 'home' %}">
+        {% trans 'Back' %}
```
ushkarev, line 15: just remembered: for this link, it's actually better to do it in the view as `context['breadcrumbs_back'] = ???` so that it appears in the breadcrumb navigation bar rather than as part of the main page contents

xoen:
@ushkarev sorry for the noise, I've rebased (although I suspect it will not be the last rebase considering the traffic in this repository these days) - and [the amendments are in this commit](https://github.com/ministryofjustice/money-to-prisoners-cashbook/pull/540/commits/247a38a9aa0a20e6170696845556ead38bc2c498) so it should be easier for you to review.

ushkarev (dismissed):

PR #541, opened 2021-03-11
Remove banners announcing November 2020 policy changes

ushkarev:
… as it's been 4 months so all users will have seen them. We don't want to distract people from other, more important, messages that we may wish to show instead.
The page they linked to remains in case it's been linked to and in case we may want to refer users in future.


xoen (approved):
👍

PR #542, opened 2021-03-12
MTP-1805 request an account

Staberinde:
Jira: https://dsdmoj.atlassian.net/browse/MTP-1805

Linked PRs:
* Common: https://github.com/ministryofjustice/money-to-prisoners-common/pull/412
* API: https://github.com/ministryofjustice/money-to-prisoners-api/pull/642

### DIFF
--- a/mtp_cashbook/apps/mtp_auth/forms.py
+++ b/mtp_cashbook/apps/mtp_auth/forms.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 import logging
 
 from django import forms
+from django.conf import settings
 from django.core.cache import cache
 from django.utils.translation import gettext_lazy as _
 from mtp_common.api import retrieve_all_pages_for_path
@@ -19,8 +20,8 @@ class CashbookSignUpForm(SignUpForm, BaseTicketForm):
     prison = forms.ChoiceField(label=_('Prison'), help_text=_('Enter the prison you’re based in'))
 
     # Non form class variables
-    cashbook_account_request_zendesk_subject = 'Requesting a new staff account - Digital Cashbook'
-    zendesk_tags = ('mtp', 'cashbook', 'account_request')
+    cashbook_account_request_zendesk_subject = 'MTP for digital team - Digital Cashbook - Requesting a new staff account'
+    zendesk_tags = ('feedback', 'mtp', 'cashbook', 'account_request', settings.ENVIRONMENT)
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -43,7 +44,6 @@ class CashbookSignUpForm(SignUpForm, BaseTicketForm):
         return choices
 
     def user_already_requested_account(self):
-        # TODO Also, is it worth adding a unique_together definition for username/role fields?
         try:
             response = self.api_session.get(
                 'requests/',
--- a/mtp_cashbook/apps/mtp_auth/forms.py
+++ b/mtp_cashbook/apps/mtp_auth/forms.py
@@ -20,7 +20,7 @@ class CashbookSignUpForm(SignUpForm, BaseTicketForm):
     prison = forms.ChoiceField(label=_('Prison'), help_text=_('Enter the prison you’re based in'))
 
     # Non form class variables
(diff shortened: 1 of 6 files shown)


mtp_cashbook/apps/mtp_auth/forms.py
```diff
@@ -7,31 +8,63 @@
...
 ACCOUNT_REQUEST_ROLE = 'prison-clerk'
 
 
-class CashbookSignUpForm(SignUpForm):
+class CashbookSignUpForm(SignUpForm, BaseTicketForm):
     prison = forms.ChoiceField(label=_('Prison'), help_text=_('Enter the prison you’re based in'))
 
+    # Non form class variables
+    cashbook_account_request_zendesk_subject = 'Requesting a new staff account - Digital Cashbook'
+    zendesk_tags = ('mtp', 'cashbook', 'account_request')
```
ushkarev, line 23: could we also add `settings.ENVIRONMENT` like in https://github.com/ministryofjustice/money-to-prisoners-cashbook/blob/b62fcd3922481b651c3d5e303a8b20efc3dc6057/mtp_cashbook/apps/feedback/views.py#L13 so that hosting environments can be distinguished?

ushkarev: i'm unsure about the 'feedback' tag cos it's on all our tickets currently… which means it's actually not useful for grouping tickets in any way.

mtp_cashbook/apps/mtp_auth/forms.py
```diff
@@ -7,31 +8,63 @@
...
                 choices = retrieve_all_pages_for_path(self.api_session, '/prisons/', exclude_empty_prisons=True)
-                choices = [[prison['nomis_id'], prison['name']] for prison in choices]
+                choices = OrderedDict([[prison['nomis_id'], prison['name']] for prison in choices])
                 cache.set('sign-up-prisons', choices, timeout=60 * 60 * 6)
             except (RequestException, OAuth2Error, ValueError):
                 logger.exception('Could not look up prison list')
         return choices
+
+    def user_already_requested_account(self):
+        # TODO Also, is it worth adding a unique_together definition for username/role fields?
```
ushkarev, line 46: do you mean on the model? i _think_ maybe it's best to not cos users shouldn't be allowed access to multiple apps (which is a slightly weird limitation)

mtp_cashbook/apps/mtp_auth/forms.py
```diff
@@ -7,31 +8,63 @@
...
 
 ACCOUNT_REQUEST_ROLE = 'prison-clerk'
 
 
-class CashbookSignUpForm(SignUpForm):
+class CashbookSignUpForm(SignUpForm, BaseTicketForm):
     prison = forms.ChoiceField(label=_('Prison'), help_text=_('Enter the prison you’re based in'))
 
+    # Non form class variables
+    cashbook_account_request_zendesk_subject = 'Requesting a new staff account - Digital Cashbook'
```
ushkarev, line 22: for the other tickets, we use the subject line to also say which team is supposed to answer it:
https://github.com/ministryofjustice/money-to-prisoners-cashbook/blob/b62fcd3922481b651c3d5e303a8b20efc3dc6057/mtp_cashbook/apps/feedback/views.py#L12
vs
https://github.com/ministryofjustice/money-to-prisoners-send-money/blob/44138128836114592a99bc347fc3573c142d2773/mtp_send_money/apps/help_area/views.py#L27
so you think it's worth saying something similar?

ushkarev (approved):

mtp_cashbook/apps/mtp_auth/forms.py
Staberinde: Happy to leave it for now, but my take is that adding that constraint (and an appropriate error message) would mean that a user can't request access to multiple apps at the same time.

mtp_cashbook/apps/mtp_auth/forms.py
Staberinde: I agree with your point about the feedback tag, i'll add here anyway for consistency

ushkarev (approved):

PR #544, opened 2021-03-17
Bundle of fixes and updates related to GOV.UK Design System update

ushkarev:
Largely updates upstream dependencies, but also includes new button group visual component (buttons placed side-by-side) from GOV.UK design system and better print layouts.


xoen (approved):
LGTM 👍

PR #545, opened 2021-03-23
MTP-1838: Simplify 'Contact us' form

xoen:
- Remove lead paragraph
- Changed textarea label to 'Please give details'

Also
- Made 'Your email address' field mandatory.

Ticket: https://dsdmoj.atlassian.net/browse/MTP-1838

**NOTE**: As part of this ticket we're reviewing the exact copy of the pre-populated text which comes from the FAQ. This is independent from this change set and I could either add to this PR later on or make that change in a separate PR.


ushkarev (approved):

xoen:
Thanks @ushkarev

PR #547, opened 2021-03-25
Remove incorrect information about user-cancelled disbursements

ushkarev:
…because they are _not_ visible in history section (under “Payments made”) and have not been for a while.


xoen (approved):
LGTM 👍

PR #561, opened 2021-08-18
Replace usage of Mailgun with GOV.UK Notify [MTP-1901]

ushkarev:


mtp_cashbook/apps/cashbook/tests/test_views.py
```diff
@@ -227,14 +232,20 @@ def test_new_credits_submit(self, mock_create_transaction):
...
-            self.assertTrue(
-                '£52.00' in mail.outbox[0].body and '£45.00' in mail.outbox[1].body or
-                '£52.00' in mail.outbox[1].body and '£45.00' in mail.outbox[0].body
+            self.assertEqual(len(mail.outbox), 0)
+            self.assertSetEqual(
+                set(
+                    data['personalisation']['amount']
+                    for data in rsps.send_email_request_data
+                ),
+                {'£52.00', '£45.00'},
```
xoen, line 241: Oh I didn't know you could initialise sets using curly braces 👍

mtp_cashbook/apps/cashbook/tests/test_views.py
```diff
@@ -570,7 +584,8 @@ def test_new_credits_submit_with_uncreditable(self, *_):
             self.assertEqual(response.status_code, 200)
             self.assertContains(response, '1 credit sent to NOMIS')
             self.assertContains(response, '1 credit needs your manual input in NOMIS')
-            self.assertEqual(len(mail.outbox), 1)
+            self.assertEqual(len(mail.outbox), 0)
```
xoen, line 587: Just to be sure I understand this correctly: `mail.outbox` had one element because it was using a Django email backend before - It has now 0 elements because emails are sent using an API call to Notify (we test the email was sent with the assertion below).

xoen (approved):
LGTM 👍

mtp_cashbook/apps/cashbook/tests/test_views.py
ushkarev: it's probably a silly thing to check, but yeah the idea was to make sure no email was sent using the django email backend

xoen (approved):
👍

xoen (approved):

xoen (dismissed):

xoen (approved):

PR #585, opened 2022-07-20
MTP-1925: Clearing a form is lost when changing pages

xoen:
Clear a form default date means that query parameter is set blank.
Unfortunately these blank query params were ignored (we think this may be
to make the URL look cleaner)

Removing the logic that removes `None` query params should fix this issue.


ushkarev (approved):
worth playing around with the forms on test just to be sure

PR #612, opened 2025-02-26
Django 4.2

xoen:
Changes needed to upgrade to Django 4.x (currently 3.2 is in use).

NOTE: So far this is not using Django 4.x but these are changes required to upgrade (e.g. because something deprecated was removed)


mtp_cashbook/apps/disbursements/views.py
```diff
@@ -216,7 +216,8 @@ def get_nomis_balances(self):
             pass
 
     def get(self, request, **kwargs):
-        if request.is_ajax():
+        is_ajax = request.headers.get('x-requested-with') == 'XMLHttpRequest'
```
ushkarev, line 219: if they removed it, does that mean it doesn't work in modern browsers anymore? this functionality may need to change; like perhaps there needs to be a separate view for the ajax-y version…

xoen:
Superseded by https://github.com/ministryofjustice/money-to-prisoners-cashbook/pull/614

